PR #10, opened 2021-01-29
30685 usagi properties file

simon-tarr:
I've created a simple test method for Usagi CLI. In this update I have:

- Created a Usagi folder containing a stand-alone version of the compiled Usagi.jar file (which outputs all fields)
- Added default-jre to dockerfile to install Java for Usagi.jar

Build and run the app, then `docker-compose exec api python manage.py test`, you will get an error that Uasgi can't find the properties file. This is expected behaviour - the error means that Usagi runs. I haven't written the code yet to perform a full run of Usagi because we need to have a discussion about where to store indexes during development (I'm not sure what the best thing to do is, because the index is large). Once we've discussed this, I can work out a way of running a full Usagi run (hard-coded) before moving onto representing the properties file in a model (to be called dynamically).

### DIFF
--- a/api/mapping/models.py
+++ b/api/mapping/models.py
@@ -57,6 +57,46 @@ class ScanReportValue(BaseModel):
     value = models.CharField(max_length=32)
     frequency = models.IntegerField()
 
+# class UsagiProperties(BaseModel):
+#     """
+#     Model to hold information on the Usagi input settings
+#     """
+#     scan_report_field = models.ForeignKey(ScanReportField, on_delete=models.CASCADE)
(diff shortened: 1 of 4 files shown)


api/mapping/models.py
```diff
@@ -57,6 +57,46 @@ class ScanReportValue(BaseModel):
     value = models.CharField(max_length=32)
     frequency = models.IntegerField()
 
+# class UsagiProperties(BaseModel):
```
simon-tarr, line 60: Added a skeleton model structure for Usagi properties and to store Usagi results. This needs a lot of work, but it's a start.

PR #17, opened 2021-02-03
29519 structural mapping form

simon-tarr:
Attempting to move the structural mapping edit form into it's own place (when you click the link 'Edit Structural Mapping'). 

I've got stuck at the final hurdle - as always it's with URL routing! I really can't work out how urls.py works with the templates and views we create. I think perhaps the code needs to be more verbose when using view classes? Can we consider using arguments in CBVs if they're available (rather than letting Django work behind the scenes)? 

For example:
```
class MyListView(ListView):
    model = MyModel
    template = "app/my_template.html"
```

Rather than:
```
class MyListView(ListView)
    model = MyModel
    ...
    <<Django magic happens here - how does the view route to our template?>>
    ...
```

Sometimes too much abstraction is a bad thing for understanding how the code works!

### DIFF
--- a/api/mapping/models.py
+++ b/api/mapping/models.py
@@ -42,8 +42,6 @@ class Mapping(BaseModel):
     def __str__(self):
         return f'{self.table, self.field}'
 
-
-
 class ClassificationSystem(BaseModel):
     """
     Class for 'classification system', i.e. SNOMED or ICD-10 etc.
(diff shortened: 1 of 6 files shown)


calmacx:
@simon-tarr, @vpnu  to avoid spamming teams..  i checked this commit out to have a play with django-select2

image

I'm not seeing any options pop up, and I started with a refresh (delete `db.sqlite` and `migrations/00*py`  etc )

vpnu (approved):

PR #25, opened 2021-02-08
Formatted some of the table entries for patient ID and date/event

fierytortilla:


simon-tarr (approved):
Thanks @fierytortilla

PR #28, opened 2021-02-09
Showall structural mappings

simon-tarr:
@calmacx I think this PR now replaces your old one due to my merge (but double check that all your changes have come through as expected).


api/mapping/views.py
```diff
@@ -100,11 +103,39 @@ class ScanReportFieldUpdateView(UpdateView):
     def get_success_url(self):
         return "{}?search={}".format(reverse('fields'), self.object.scan_report_table.id)
 
+
+class ScanReportValueUpdateView(UpdateView):
```
calmacx, line 107: this is redundant, let me remove to avoid confusion

calmacx (changes requested):
* [ ] remove redundant edit values - Calum

calmacx (approved):

PR #40, opened 2021-02-19
Tidy webapp breadcrumbs

vpnu:
This PR includes:

- Adding breadcrumbs to documents file status page
- Changing navigation bar colours to match the co-connect site
- Add bootstrap to password change form


api/templates/base.html
```diff
@@ -14,7 +14,7 @@
     <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap-icons@1.3.0/font/bootstrap-icons.css">
 
     <!-- our custom css -->
-    <link rel="stylesheet" href="{% static 'style/custom.css' %}">
+    <link rel="stylesheet" href="{% static 'style/custom.css'  %}?rnd=132">
```
calmacx, line 17: what does this do? More for my own interest..

api/templates/registration/password_change_form.html
```diff
@@ -10,9 +10,22 @@
 </nav>
 
   <h2>Change password</h2>
-  <form method="post">
-    {% csrf_token %}
-    {{ form.as_p }}
-    <button type="submit">Change password</button>
+  <script src="//cdnjs.cloudflare.com/ajax/libs/jquery/3.2.1/jquery.min.js"></script>
```
calmacx, line 13: how come ajax include is added here? Should be in `base.html` template already?

calmacx (approved):
Probably best to remove the addition of including jquery again, should make a difference as should already be include (unless ive missed something), so approving anyway

api/templates/base.html
vpnu: While I was changing the .css file the changes wouldn't show on the site immediately so I added this a random version number at the end so it would refresh every time

api/templates/registration/password_change_form.html
vpnu: Yeah I think you're right theres no need for it to be here so I'll just remove it

api/templates/base.html
calmacx: ohh cool, i just keep hitting command+shift+r on my mac to get the stylesheets to reload ha!

PR #49, opened 2021-03-02
30683 usagi input

simon-tarr:
Apologies for the big PR! This is a large update which implements a number of core updates to the webapp. I did try and break these down into smaller parts (for smaller PRs) but there was just too much interdependencies to make this viable. So, apologies there's a lot to review here!

I am merging into master now so that I can start to distribute a number of smaller tasks to improve the UI and functionality. The functionality here by no means complete and there is a lot of UI/UX work that we'll need to complete before this is ready for prime time.

At it's core this update allows:

- The creation of what I'm calling a 'skeleton' data dictionary. This creates a bare-bones dictionary when a user _uploads a new scan report_ (it's essentially based on the field/value combinations from the scan report)
- Using Vas' documents section, you can upload an external data dictionary to the system. Set the dictionary to 'Live' for later code to work (more on this shortly). Make sure your scan report and data dictionary belong to the same data provider. The code works by finding a 'Live' dictionary in the files section and merging this into the data dictionary for that Data Partner's scan report. The tool won't work if you have a Panther Scan report but a data dictionary associated with, say, Match.
- Make a note of the Scan Report PK and navigate to http://localhost:8080/datadictionary/merge/?search=**_INSERTSCANREPORTPKHERE_** to kick off the merge. You can see it's status in the console.
- Once completed if you navigate to the scan report's data dictionary page, you will see that the external data has been merged into the internal ('skeleton' dictionary).

As a side note, I think some of the code changes are simple formatting changes. Python Black for VSCode seems to have done a little formatting in places which Black in PyCharm wouldn't have done. I'm not sure why the style guidelines are different between the two implementations. Anyway, those changes won't affect the code's functionality, just how it looks.

There are also some Usagi-based changes. These changes are just initial changes which we can develop upon. Now the above data dictionary code is working, we can pass in data to Usagi. At present, you won't be able to test that Usagi is working correctly (due to Java/Docker issues)


api/mapping/services.py
```diff
@@ -174,26 +179,97 @@ def process_scan_report(scan_report_id):
...
-    filepath = filepath
 
-    with open(filepath, 'rt') as f:
-        reader = csv.reader(f)
-        next(reader)  # Skip header row
+def build_usagi_index():
+    # Use 'build' to create index for the first time.
+    # Index stored in data/usagi/mainIndex
+    p = subprocess.Popen(
+        "java -jar Usagi.jar build usagi_build_index.properties",
```
calmacx, line 192: note to self: check Usagi.jar can be found

api/mapping/services.py
```diff
@@ -174,26 +179,97 @@ def process_scan_report(scan_report_id):
...
+        "Value",
+        "Frequency",
+        "FieldDesc",
+        "ValueDescription",
+    ]
+
+    df = dict_df
+
+    # df = df.head(10)
+    df.to_csv("/data/usagi/input/usagi_input_data.csv", index=False)
```
calmacx, line 256: What's the likelihood of two instances of Usagi being run at the same time by different users?
I.e. would it be safer to have the report name/id in the temp save file so there can't be overwriting issues?

Suggestion:
```
df.to_csv(f"/data/usagi/input/usagi_input_data_{scan_report_id}.csv", index=False)
```

api/mapping/test_services.py
```diff
@@ -20,3 +20,15 @@ def test_process_scan_report_sheet_table(self):
...
+    # Run this to build Usagi's index for the first time
+    # Takes a long time to run!
+    def test_build_usagi_index(self):
+        x = build_usagi_index()
+        print(x)
+
+    # Run this to get Usagi going
+    def test_run_usagi(self):
+        x = run_usagi()
+        print(x)
```
calmacx, line 34: not clear to me if these are needed - but we can leave them in

api/mapping/urls.py
```diff
@@ -7,16 +7,23 @@
...
+    path('tables/', login_required(views.ScanReportTableListView.as_view()), name='tables'),
+    path('fields/', login_required(views.ScanReportFieldListView.as_view()), name='fields'),
+    path('fields/<int:pk>/update/', login_required(views.ScanReportFieldUpdateView.as_view()), name='scan-report-field-update'),
+    path('fields/<int:pk>/create_mapping/', login_required(views.AddMappingRuleFormView.as_view()), name='create-mapping-form'),
+    path('fields/<int:pk>/mapping_rules/', login_required(views.StructuralMappingListView.as_view()), name='view-structural-mapping'),
+    path('fields/<int:pk>/mapping_rules/delete', login_required(StructuralMappingDeleteView.as_view()), name='structural-mapping-delete'),
+    path('values/', login_required(views.ScanReportValueListView.as_view()), name='values'),
+    path('scanreports/', login_required(views.ScanReportListView.as_view()), name='scan-report-list'),
+    path('scanreports/<int:pk>/mapping_rules', login_required(views.StructuralMappingTableListView.as_view()), name='tables-structural-mapping'),
+    path('scanreports/create/', login_required(views.ScanReportFormView.as_view()), name='scan-report-form'),
```
calmacx, line 19: there might have been a master merge issue here?

I believe `login_required` got removed by @vpnu and instead we used `@login` decorator in `view.py`

api/mapping/views.py
```diff
@@ -161,45 +191,45 @@ def get_context_data(self, **kwargs):
...
-    template_name = 'mapping/mappingrule_form.html'
+    template_name = "mapping/mappingrule_form.html"
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
-        scan_report_field = ScanReportField.objects.get(
-            pk=self.kwargs.get('pk')
-        )
+        scan_report_field = ScanReportField.objects.get(pk=self.kwargs.get("pk"))
```
calmacx, line 214: Just a quick comment, there seems to be a lot of changes like this
* `'` --> `"` 
* multi-line functions moved to single-line

Not sure if it's you making these changes, or it's come from a bad merge to the master where these changes are being overwritten.

The PEP8 style guide says to use the multiline split, make it easier to read
i.e.
```
 scan_report_field = ScanReportField.objects.get(
            pk=self.kwargs.get('pk')
        )
```


Now this is super pedantic of me to impose PEP8 style guides 😂  - im just wondering why these style changes are coming in with this MR? And making sure it's not something that's happening because of a merge to master, where someone else has made these changes?

api/mapping/views.py
```diff
@@ -210,56 +240,62 @@ def form_valid(self, form):
...
     def get_success_url(self):
-        scan_report_field = ScanReportField.objects.get(
-            pk=self.kwargs.get('pk')
-        )
+        scan_report_field = ScanReportField.objects.get(pk=self.kwargs.get("pk"))
 
-        return "{}?search={}".format(reverse('fields'), scan_report_field.scan_report_table.id)
+        return "{}?search={}".format(
+            reverse("fields"), scan_report_field.scan_report_table.id
+        )
```
calmacx, line 247: another example of random changing of single line to multiple lines

api/mapping/views.py
```diff
@@ -545,35 +646,171 @@ def password_reset_request(request):
...
+        "DataSet",
+        "Table",
+        "Field",
+        "Value",
+        "Frequency",
+        "FieldDesc",
+        "ValueDescription",
+    ]
+
+    dict_df.to_csv('/data/TEMP_internal_dictionary.csv')
```
calmacx, line 737: as before, double check can't get screwed up when two instances running? Add some sort of hash/id to the name?

api/mapping/views.py
```diff
@@ -545,35 +646,171 @@ def password_reset_request(request):
...
+                continue
+
+            else:
+                obj.dictionary_table=row['Dictionary_TableName']
+                obj.dictionary_field=row['Dictionary_ColumnName']
+                obj.dictionary_field_description=row['Dictionary_ColumnDesc']
+                obj.dictionary_value_code=row['Dictionary_ValueCode']
+                obj.dictionary_value_description=row['Dictionary_ValueDescription']
+                obj.save()
+
```
calmacx, line 814: happy to trust you on all this :D 
maybe i have some suggestions to simplify, but dont think this is the time

calmacx (changes requested):

api/mapping/urls.py
vpnu: Yeah it was removed a while ago right after the documents page was added

vpnu: now we just need @method_decorator(login_required,name='dispatch') before the view class

api/mapping/views.py
simon-tarr: Ah balls, forgot to take these out. The `to_csv()` calls aren't required, they were just there for debugging.

api/mapping/views.py
simon-tarr: I _think_ it might have been a botched run of Python Black in VSCode.

api/mapping/urls.py
simon-tarr: Yeah the decorator makes more sense, I don't recall `login_required` being removed though. Could well be a master merge issue.

api/mapping/views.py
simon-tarr: Yes, think this was Python Black again....

api/mapping/views.py
simon-tarr: Any suggestions welcomed. Written this super verbose so I'm crystal clear about what's being saved.

api/mapping/services.py
simon-tarr: Usagi.jar won't be present in this PR....it'll come a bit later on. At the moment, we can't run Usagi due to Docker performance issues.

api/mapping/services.py
simon-tarr: Reasonably likely I would say!

Not sure if I mentioned earlier in the Dev team meeting - this PR really is pretty ropey but I was keen to get the code into master so I can start dishing out smaller tasks to the team. If I wait until all the functionality is done, the merge would be a mightmare, I think.

simon-tarr: I should add, the functionality in this PR can't actually be accessed/run by the app users (unless they know specific URLs to call the methods e.g. http://localhost:8080/datadictionary/merge/?search=1 and http://localhost:8080/testusagi/1/)

api/mapping/views.py
simon-tarr: For example, Python black formatted the import models call like:

```
from .models import (
    ScanReport,
    ScanReportTable,
    ScanReportField,
    ScanReportValue,
    DataDictionary,
)

```

Is this in line with PEP8?

api/mapping/views.py
calmacx: we can add it to a backlog in the far future to have a code review / cleanup/ part of a hackathon

api/mapping/services.py
calmacx: cool well i'll let you add is as follow up task then

api/mapping/urls.py
calmacx: Yeh @simon-tarr  i think it just needs these removing / rebase/remerge with the master and then i can approve and merge to master when that is done

calmacx (approved):

PR #53, opened 2021-03-05
Changes in CSS

vpnu:
This PR includes:
- Change the colours of the buttons to be the same as the standard co-connect palette
- Remove hover up effect from navigation bar

### DIFF
--- a/api/static/style/custom.css
+++ b/api/static/style/custom.css
@@ -1,12 +1,12 @@
 /*
-  == primary pallete == 
+  == primary palette == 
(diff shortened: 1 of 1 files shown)


calmacx (approved):
Just a quick css tip. The palette colours are defined in `root:`

So instead of
```
 background-color: #475da7;
```
You can do:
```
background-color: var( --co-connect-primary);
```

Maybe you know this already, hope this doesnt sound like i am mansplaining 😄!!

No need to change or anything.

PR #55, opened 2021-03-09
31851 check data dictionary column names

fierytortilla:
Hey so the code is set up in forms.py. There needs to be some refactoring done there so the code doesn't repeat itself in the DocumentForm and DocumentFileForm classes, but otherwise, it works fine.


api/mapping/forms.py
```diff
@@ -123,17 +123,44 @@ class DocumentForm(forms.Form):
...
-    document = forms.ModelChoiceField(label="Document", queryset=Document.objects.all())
+    document = forms.ModelChoiceField(
+        label="Document", queryset=Document.objects.all()
+    )
     description = forms.CharField(
         label="Document Description",
         widget=forms.TextInput(attrs={"class": "form-control"}),
     )
 
+    # clean_document_file(forms.Form)
```
vpnu, line 152: same with this one

vpnu (approved):

simon-tarr (approved):

PR #59, opened 2021-03-17
Add conceptID in the ScanReportField model

vpnu:
This PR:

- Adds concept_id as a field in ScanReportField model and sets the default value to be -1
- Allows the user to enter a concept_id for each column if needed and save the entry to the model


fierytortilla (approved):
:)

PR #61, opened 2021-03-18
Revert "Add conceptID in the ScanReportField model"

calmacx:
Reverts CO-CONNECT/mapping-pipeline#59

### DIFF
--- a/api/mapping/models.py
+++ b/api/mapping/models.py
@@ -170,7 +170,8 @@ class ScanReportField(BaseModel):
         null=True,
         blank=True
     )
-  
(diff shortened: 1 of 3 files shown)


calmacx:
Morning @vpnu @fierytortilla , need to revert #59  as it shouldn't have been approved as it is.
Main problem is:
```
concept_id = models.IntegerField(default=-1) 
```
Needs to be allowed to be null/blank or it causes crashes. Unless we've made the decision to delete everything in the db and start from scratch.

fierytortilla (approved):
:-)

PR #62, opened 2021-03-23
32420 import classification system from scan report

fierytortilla:
Screenshot 2021-03-21 at 20 11 41
When the data team modifies the White Rabbit scan reports as depicted in the uploaded screenshot, they can include the classification system of each source field which can be incorporated into the site's scan report.


simon-tarr:
@fierytortilla could you merge master with your branch (PR) and resolve the conflicts before we review? Cheers!

calmacx (approved):

PR #64, opened 2021-03-24
The ability for users to structural mapping has been removed

fierytortilla:
Besides removing the HTML page and view class designed for adding structural mapping, I have also deleted the deletion HTML pages and view class. Let me know if the deletion functionality is still required, or if I removed anything incorrectly in any way.


vpnu:
There's a quite a few changes missing from this David. We also need to remove the model from admin.py, comment the code for the model in models.py remove MappingRule from import statements at the top etc. So I'm going to go ahead and close this PR

PR #66, opened 2021-03-25
32961/make json work again

calmacx:


api/mapping/templates/mapping/scanreportfield_list.html
```diff
@@ -48,18 +47,7 @@ <h2>Fields</h2>
 		  {{ object.classification_system.name }}
 		</td>
 
-                <td>
-                    {% if object.is_patient_id %}
```
vpnu, line 52: how come we're removing this one? we don't need it anymore?

vpnu (approved):

PR #67, opened 2021-03-29
Update/clean mapping table view

calmacx:
image


vpnu:
Do we need to merge into co-connect-tools master as well before merging this? since your change in omop_db_inspect is pointing to co-connect-tools master?

calmacx:
yes please @vasilikipn , not sure if that change will screw up anything else you are doing though?

vpnu (approved):

PR #72, opened 2021-03-31
32450/create assertions page

vpnu:
This PR:

- Creates a new model called ScanReportAssertions, where we store positive and negative assertions for each scan report
- Allows the user to add assertions
- Allows the user to edit the assertions


simon-tarr:
@calmacx Could you review this for me please? Since I've commited to the PR I can no longer review!
In addition to Vas' comments above, I'm simply using the data in the new model to filter the data dictionary list.

simon-tarr:
Forgot to say - the new model will necessitate migrations (although I suspect that's pretty obvious...)

calmacx (approved):

PR #91, opened 2021-04-21
NLP Integration

simon-tarr:
This PR allows users to submit a string to the NLP service and it will return SignsOrSymptoms for that string. 

There's a new model in this PR which will require migrations but the model is 'independent' (i.e. it isn't linked to any other exant tables in the app). Therefore we can easily remove this model at a later date if we no longer want to provide basic NLP lookup to users.


api/mapping/admin.py
```diff
@@ -1,10 +1,23 @@
 from django.contrib import admin
 
-from .models import DocumentType, Source, DataPartner, ScanReport, \
-    ScanReportTable, \
-    ScanReportField, ScanReportValue, ScanReportAssertion, \
-    ClassificationSystem, OmopTable, OmopField, Document, \
-    DocumentFile, DataDictionary, StructuralMappingRule
+from .models import (
```
simon-tarr, line 3: Formatted with PEP8

vpnu, line 3: is this after isort?

simon-tarr: Nope, just PEP8. I _think_ isort also puts imports into alphabetical order.

vpnu (approved):

PR #94, opened 2021-04-22
Update/scan report upload

vpnu:
This PR is for:

- Remove old checks for pass_from_source and ignore flag from process_scan_report

- Checking that the Scan Report has the correct column names/content before uploading on the site.

Screenshot 2021-04-22 at 14 27 18
Screenshot 2021-04-22 at 14 27 40

- When someone adds a conceptID on a column level, disable the link to the values to prevent mapping on both levels:
Screenshot 2021-04-22 at 15 28 13


simon-tarr:
@vasilikipn Do you think it's worth setting the error message on a different colour background, like we do with the table filters on the [mapping rules page](http://localhost:8080/scanreports/9/mapping_rules/)?

vpnu:
> @vasilikipn Do you think it's worth setting the error message on a different colour background, like we do with the table filters on the [mapping rules page](http://localhost:8080/scanreports/9/mapping_rules/)?
sure like a yellow or red?

api/mapping/services.py
```diff
@@ -146,16 +146,6 @@ def process_scan_report(scan_report_id):
                     else:
                         scanreport.is_date_event = False
 
-                    if scanreport.ignore_column == "Ignore":
```
simon-tarr, line 149: We still need this logic to build a more concise data dictionary for NLP. Can you reinstate this code, please? Thanks!

simon-tarr (commented):
Just need to reinstate the code in services.py

simon-tarr (approved):

PR #100, opened 2021-04-28
#AB34487 Added ScanReportConceptID model

simon-tarr:
This PR adds a new model for NLP conceptIDs

### DIFF
--- a/api/mapping/models.py
+++ b/api/mapping/models.py
@@ -550,7 +550,7 @@ class NLPModel(models.Model):
         return str(self.id)
 
 
-class ScanReportConceptID(BaseModel):
+class ScanReportConcept(BaseModel):
     concept_name = models.CharField(max_length=128)
     concept_id = models.CharField(max_length=16)
     entity = models.CharField(max_length=64)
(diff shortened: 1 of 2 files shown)


api/mapping/models.py
```diff
@@ -535,22 +528,40 @@ class DataDictionary(BaseModel):
...
         blank=True,
         null=True,
     )
-    
+
+    def __str__(self):
+        return str(self.id)
+
+
+class ScanReportConceptID(BaseModel):
```
py5gol, line 553: Can we call it just ScanReportConcept?

api/mapping/models.py
```diff
@@ -535,22 +528,40 @@ class DataDictionary(BaseModel):
...
+    concept_name = models.CharField(max_length=128)
```
py5gol, line 554: With this, and many other in the models.py, could you break break the model attributes in multiple lines?

py5gol (changes requested):

api/mapping/models.py
simon-tarr: @py5gol I can do, yes. But my implementation of PEP8 only automatically breaks onto multiple lines when the line length gets to a certain length (not sure about the number of characters though)

simon-tarr: I think it also splits onto multiple lines if number of attributes >2

vpnu (approved):

py5gol (approved):

PR #106, opened 2021-05-06
Remove containe db from docker-compose.yml. Remove OMOPDetails call o…

py5gol:
…n views.py. Tidy up views.py imports.

### DIFF
--- a/api/mapping/views.py
+++ b/api/mapping/views.py
@@ -1,27 +1,21 @@
 import ast
 import json
-import os
-import sys
-from io import BytesIO, StringIO
+from io import StringIO
 
-import coconnect
-import pandas as pd
 from coconnect.tools import dag, mapping_pipeline_helpers
-from coconnect.tools.omop_db_inspect import OMOPDetails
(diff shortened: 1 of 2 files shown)


api/mapping/views.py
```diff
@@ -50,28 +44,20 @@
...
 from .tasks import (
     nlp_single_string_task,
     process_scan_report_task,
 )
 
-# to refresh/resync with loading from the database, switch to:
-# omop_lookup = OMOPDetails(load_from_db=True)
-# this will take longer, but it will recreate the csv dump of all the
-# omop fields
-omop_lookup = OMOPDetails()
```
vpnu, line 73: Wouldn't this cause the mapping rule generation to break? Since we're still using it in def generate? In this line rules_set = omop_lookup.get_rules(concepts)

py5gol: Thanks @vpnu, I have reintroduced it inside the class based view.

vpnu (approved):

PR #108, opened 2021-05-06
add db compose only file

calmacx:


py5gol (changes requested):
Hi @Calum, I am happy with this going back into the main docker-compose.py and each dev configuring the .env file as needed.

PR #109, opened 2021-05-06
Update the values page to allow multiple concept ids per scanreportvalue

py5gol:


api/mapping/forms.py
```diff
@@ -195,4 +195,13 @@ class ScanReportAssertionForm(forms.Form):
 class NLPForm(forms.Form):
     user_string = forms.CharField(
         label = ""
-    )
\ No newline at end of file
+    )
+
+class ScanReportValueConceptForm(forms.Form):
+    scan_report_value_id=forms.IntegerField(
```
simon-tarr, line 201: Do we want to keep fields with no arguments on a single line?

api/mapping/forms.py
```diff
@@ -195,4 +195,13 @@ class ScanReportAssertionForm(forms.Form):
...
+
+    )
+
+    concept_id=forms.IntegerField(
```
simon-tarr, line 205: Same here.

api/mapping/models.py
```diff
@@ -211,6 +213,31 @@ def __str__(self):
         return str(self.id)
 
 
+class ScanReportConcept(BaseModel):
```
simon-tarr, line 216: The only field "missing" here is a field to record the string we send to NLP:

`processed_string = models.CharField(max_length=64)`

api/mapping/views.py
```diff
@@ -1147,3 +1155,54 @@ def get_context_data(self, **kwargs):
...
+                )
+            except Concept.DoesNotExist:
+                messages.error(request,
+                                 "Concept id {} does not exist in our database.".format(form.cleaned_data['concept_id']))
+                return redirect("/values/?search={}".format(scan_report_value.scan_report_field.id))
+
+            scan_report_concept = ScanReportConcept.objects.create(
+                concept_name=concept.concept_name,
+                concept_id=concept.concept_id,
+                entity="empty",
```
simon-tarr, line 1181: Do we want to hardcode this as "empty" or leave the field null/blank?

api/mapping/views.py
```diff
@@ -1147,3 +1155,54 @@ def get_context_data(self, **kwargs):
...
+                entity_type="empty",
```
simon-tarr, line 1182: Same here.

api/mapping/views.py
```diff
@@ -1147,3 +1155,54 @@ def get_context_data(self, **kwargs):
...
+                confidence=1.0,
+                vocabulary="empty",
```
simon-tarr, line 1184: And for the next two lines.

simon-tarr (changes requested):
I've left a few comments for you @py5gol. Nothing necessarily stopping the merge, but a few points of clarification.

api/mapping/forms.py
py5gol: I like this approach because if a parameter is added in the future, it will involve one change, instead of 3 (one substitution and two additions).

api/mapping/forms.py
py5gol: Same above.

api/mapping/views.py
py5gol: Same above.

api/mapping/views.py
py5gol: Good point. I am not sure we need this attribute in the model. How are we using it?

api/mapping/views.py
py5gol: Same above.

api/mapping/models.py
py5gol: I've added it, but I think we might need to come back to it later and discuss a better place to store this information.

api/mapping/views.py
simon-tarr: These fields are used to save the return from NLP. e.g. 'entity' could be DrugAdministered or SignOrSymptom.

api/mapping/forms.py
simon-tarr: @py5gol Personally, I don't see what the issue is of 1 change versus 3 (from a Git perspective).
I think the code is more readable if this is on a single line - I consider readabiloty more important than whether Git cosiders it 1 or 3 changes. Just my 2c anyway!

simon-tarr (approved):

PR #112, opened 2021-05-07
All ScanReportFields are now set to 'Pass from Source' by default

simon-tarr:
This PR sets all ScanReportFields to "Pass from Source" by default (on scan report upload). This ultimately changes the number of items in the data dictionary list view (and therefore what data is sent to NLP).

This change has been implemented because it's probably better to allow users to select which field values they should let through, rather than having them filter out everything that they don't want to send to NLP (i.e. is quick to let data through than it is to filter out everything that isn't required)


api/mapping/services.py
```diff
@@ -133,7 +133,7 @@ def process_scan_report(scan_report_id):
                         is_patient_id=False,
                         is_date_event=False,
                         is_ignore=False,
-                        pass_from_source=False,
+                        pass_from_source=True,
```
py5gol, line 136: @simon-tarr, could you set pass_from_source default equals True in the models instead of here. It will make sure that we got this rule independent from where we create the object.

py5gol (changes requested):

simon-tarr:
Hi @py5gol, set pass_from_source=True at the model level as requested. I've also merged master with this branch so you should be good to merge if you're happy with the latest change (y)

py5gol (approved):

py5gol (approved):

PR #113, opened 2021-05-11
34487/save nlp results

simon-tarr:
This PR pulls togther a lot of the NLP code I've been working on for the last week or so.
It's purpose is just to get this code into master so that I can branch again and further refine the NLP code.


api/mapping/views.py
```diff
@@ -1195,6 +1208,259 @@ def get_context_data(self, **kwargs):
             context = {"user_string": query.user_string, "results": json_response}
             return context
 
+def run_nlp(request):
+
```
py5gol, line 1212: Any chance you can wrap it around a function in nlp_services?

py5gol (changes requested):

api/mapping/views.py
simon-tarr: I was actually going to do this when I was still working on the "old" version on the NLP code. Do you want this to happen for this PR? Or shall I do it once we've merged and I've branched off master again?

py5gol (approved):

PR #114, opened 2021-05-11
34905/rework structural mapping

calmacx:


api/mapping/models.py
```diff
@@ -255,6 +255,17 @@ class ScanReportConcept(BaseModel):
         blank=True,
     )
 
+    #allow this to be null
```
py5gol, line 258: What is the purpose of source_concept?

api/mapping/models.py
```diff
@@ -499,32 +510,31 @@ class StructuralMappingRule(BaseModel):
     """
     To come
     """
-
+    #save the scan_report link to make it easier
     scan_report = models.ForeignKey(ScanReport, on_delete=models.CASCADE)
 
     omop_field = models.ForeignKey(OmopField, on_delete=models.CASCADE)
 
-    source_table = models.ForeignKey(
```
py5gol, line 507: Can we keep this fields for now, and add a TODO to stop using it. It will simplify testing this code without much changes in the models.

api/mapping/models.py
```diff
@@ -499,32 +510,31 @@ class StructuralMappingRule(BaseModel):
...
-        ScanReportTable, on_delete=models.CASCADE, blank=True, null=True
-    )
+    source_field = models.ForeignKey(ScanReportField,
+                                     on_delete=models.CASCADE,
+                                     null=True,
+                                     blank=True)
 
-    source_field = models.ForeignKey(
```
py5gol, line 511: Same above.

api/mapping/models.py
```diff
@@ -499,32 +510,31 @@ class StructuralMappingRule(BaseModel):
...
-        ScanReportField,
-        on_delete=models.CASCADE,
-        blank=True,
-        null=True
-        # limit_choices_to= {'scan_report_table': source_table}
-    )
+    #term_mapping = models.JSONField(null=True,blank=True)
+    do_term_mapping = models.BooleanField(default=False)
 
-    term_mapping = models.CharField(max_length=10000, blank=True, null=True)
```
py5gol, line 519: Same above.

api/mapping/models.py
```diff
@@ -499,32 +510,31 @@ class StructuralMappingRule(BaseModel):
...
-
-    operation = models.CharField(
```
py5gol, line 521: Same above.

api/mapping/models.py
```diff
@@ -499,32 +510,31 @@ class StructuralMappingRule(BaseModel):
...
 
-    source_field = models.ForeignKey(
-        ScanReportField,
-        on_delete=models.CASCADE,
-        blank=True,
-        null=True
-        # limit_choices_to= {'scan_report_table': source_table}
-    )
+    #term_mapping = models.JSONField(null=True,blank=True)
+    do_term_mapping = models.BooleanField(default=False)
```
py5gol, line 524: What does do_term_mapping do?

api/mapping/models.py
```diff
@@ -499,32 +510,31 @@ class StructuralMappingRule(BaseModel):
     """
```
py5gol, line 510: Please format this class based on the class above.

api/mapping/templates/mapping/mappingrulesscanreport_list.html
```diff
@@ -78,21 +77,35 @@ <h6>Filter On Source Table</h6>
...
         <td style="width: 16.6%">
 	  <a href="{% url 'values' %}?search={{ object.source_field.id }}"> {{object.source_field.name}} </a>
 	</td>
 	  
         <td style="width: 16.6%">
-	  {% if object.term_mapping %}
-	  {% if object.term_mapping != "null" %}
-	  {{ object.term_mapping }}
-	  {% endif %}
+	  {% if object.do_term_mapping %}
```
py5gol, line 88: Related to question above, what does do_term_mapping do here?

api/mapping/templates/mapping/scanreportfield_list.html
```diff
@@ -27,9 +27,12 @@ <h2>Fields</h2>
                 <th scope="col">Field</th>
                 <th scope="col">Data Type</th>
 		        <th scope="col">Classification System</th>
+
+	        <!---
```
py5gol, line 31: Do we want to remove these columns?

api/mapping/templates/mapping/scanreportvalue_list.html
```diff
@@ -41,12 +41,25 @@ <h2>Values</h2>
                             {{ object.frequency }}
                         </td>
                         <td>
-                            {% for concept in object.concepts.all %}
-                                <span>
-                                    {{ concept.concept.concept_id }} -
-                                    {{ concept.concept.concept_name }} ({{ concept.concept.domain_id }})
-                                </span> <a href="{% url 'scan_report_value_concept-delete' %}?scan_report_field_id={{ object.scan_report_field.id }}&scan_report_concept_id={{ concept.id }}">(delete)</a><br/>
-                            {% endfor %}
+			  <table class="table table-bordered table-sm">
```
py5gol, line 44: Is the code below assuming the user can provide a non valid, non standard concept? If so, that is not the case. When associating a new concept to a ScanReportConcept, the concept will be validated. Hence, the webapp does not need to keep track of what is the original concept.

api/mapping/views.py
```diff
@@ -63,11 +64,29 @@
 )
 
 
+m_allowed_tables = ['person','measurement','condition_occurrence','observation']
```
py5gol, line 67: Can you move it to a service, please?

api/mapping/views.py
```diff
@@ -139,6 +158,7 @@ class ScanReportFieldListView(ModelFormSetView):
     model = ScanReportField
     fields = ["is_patient_id", "date_type", "concept_id"]
     fields = ["is_patient_id", "is_birth_date", "is_date_event", "concept_id"]
+    fields = ["concept_id"]
```
py5gol, line 161: Is this needed? This class will move back to a ListView.

api/mapping/views.py
```diff
@@ -300,7 +320,7 @@ def get_context_data(self, **kwargs):
 @method_decorator(login_required, name="dispatch")
 class StructuralMappingTableListView(ModelFormSetView):
```
py5gol, line 321: Please, convert this class to a ListView, and remove the approved field from the screen. This is turning into a visualisation page, instead of a form.

api/mapping/views.py
```diff
@@ -1196,6 +1218,89 @@ def get_context_data(self, **kwargs):
             return context
 
 
+def save_mapping_rules(request,scan_report_concept):
```
py5gol, line 1221: This function could go into a service.

api/mapping/views.py
```diff
@@ -1204,23 +1309,58 @@ def save_scan_report_value_concept(request):
...
+            
             try:
-                concept = Concept.objects.get(
+                source_concept = Concept.objects.get(
                     concept_id=form.cleaned_data['concept_id']
                 )
             except Concept.DoesNotExist:
                 messages.error(request,
-                                 "Concept id {} does not exist in our database.".format(form.cleaned_data['concept_id']))
+                                 "Source Concept id {} does not exist in our database.".format(form.cleaned_data['concept_id']))
```
py5gol, line 1319: This is related to a comment above. If the concept does not exist in the Concept table, we just inform the user. Later, we might want to let the user know that the concept maps to another, but it will the user responsibility to use it.

api/mapping/views.py
```diff
@@ -1204,23 +1309,58 @@ def save_scan_report_value_concept(request):
...
+                source_concept = None
+                
             scan_report_concept = ScanReportConcept.objects.create(
+                source_concept=source_concept,
                 concept=concept,
                 content_object=scan_report_value,
             )
 
-            messages.success(request, "Concept {} - {} added successfully.".format(concept.concept_id, concept.concept_name))
+            if source_concept is None:
```
py5gol, line 1350: See comment above.

api/mapping/templates/mapping/scanreportvalue_list.html
calmacx: @py5gol this was marked as "a draft PR" so wasnt expecting or meaning for you to look at it yet.

Still, this is an important point, that would mean I could remove some unneeded features, and save me a lot of time with extra work... 
I started to play with adding this feature to save the source_concept_id as well as the concept_id because I thought it was needed, because of this example:

In Panther they mapped "4057568 - Black African (Race)" which is Non-Standard but valid (source_concept_id)
The standard relation is 38003600 - African (Race) (concept_id)

So if we/user was to enter `4057568` the validation should be rejecting it, so that we always have source_concept_id == concept_id?

api/mapping/models.py
calmacx: adding the following comment
```
# Flag if this is a rule that should be using term_mapping or not                                                  
    # - rules such as source_value should apply no term mapping                                                        
    # - we need a flag as we want to preserve a link to the ScanReportConcept                                          
    # - this is because, if that is deleted, we want to remove the rule                                                
    # - however, we dont want to ever apply any of the concept_id term_mappings                                        
    #   if this was source_value, person_id, birth_date etc.                                                           
    do_term_mapping = models.BooleanField(default=False)
```

api/mapping/models.py
calmacx: If we want to save a Non-Standard concept along side the Standard-Concept

I've added more code comments about this.

As explained, it was probably beyond the scope of my task as I had thought my task was to just get the full rule generation working again.

At the moment, this isn't used as now I implemented a rejection of Non-Standard concepts when entering concepts. Therefore this will always be null or the same Concept object as the concept

api/mapping/templates/mapping/scanreportfield_list.html
calmacx: we dont use them any more, but i'll remove and leave for another PR

api/mapping/views.py
calmacx: will leave this for a 3rd PR 👍

api/mapping/services_rules.py
```diff
@@ -1,9 +1,288 @@
...
 
 
+class NonStandardConceptMapsToSelf(Exception):
+    pass
+
+           
+#allowed tables
+m_allowed_tables = ['person','measurement','condition_occurrence','observation']
+
+#names of date objects in ScanReportTable
```
py5gol, line 14: We are rollbacking this approach. According to the new standard's doc, each table can have only one date field. So, if a table contains conditions and observations, both must use the same date column.

py5gol (changes requested):

api/mapping/models.py
```diff
@@ -495,29 +506,80 @@ def __str__(self):
         return str(self.id)
 
 
+#!! TODO --- Give this model a better name(?)
 class StructuralMappingRule(BaseModel):
```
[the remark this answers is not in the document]
calmacx: im leaving as a todo for a future task/PR, unless you want it done now?

api/mapping/models.py
```diff
@@ -495,29 +506,80 @@ def __str__(self):
...
+    #   - If this was a rule for the concept_id, use_source_concept_id=False
+    use_source_concept_id = models.BooleanField(default=False)
+
+    # Flag if this is a rule that should be using term_mapping or not
+    # - rules such as source_value should apply no term mapping
+    # - we need a flag as we want to preserve a link to the ScanReportConcept
+    # - this is because, if that is deleted, we want to remove the rule
+    # - however, we dont want to ever apply any of the concept_id term_mappings
+    #   if this was source_value, person_id, birth_date etc. 
+    do_term_mapping = models.BooleanField(default=False)
```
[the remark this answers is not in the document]
calmacx: What about the rule for <domain>_source_value? this is why i have this

api/mapping/models.py
```diff
@@ -495,29 +506,80 @@ def __str__(self):
...
+    #   associated with the mapping of a source field to a destination field
+    # - e.g. 'Y':1234 (ScanReportConcept_0), 'Yes':4321 (ScanReportConcept_1)
+    concepts = models.ManyToManyField(
+        ScanReportConcept
+    )
+
+    # optional bool field to determine if the rule uses:
+    #   - False: ScanReportConcept.concept (Standard)
+    #   - True:  ScanReportConcept.source_concept (could be Non-Standard)
+    # An example of this:
```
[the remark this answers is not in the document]
calmacx: There would be separate rules for:
* domain_concept_id
* domain_source_concept_id
* domain_source_value
* person_id
* domain_date_event

This allows us to associate the same ScanReportConcept object to all rules.  

This functionality isn't really used at the moment, but I see it as a good solution if we ever wanted to keep a source_concept as well as a concept

api/mapping/services_rules.py
calmacx: ok, but since this has not yet been rolled back, is there much I can do here?

or should it become something like this:
```
    'condition_occurrence':'observation_date',
    'observation':'observation_date'
```

api/mapping/models.py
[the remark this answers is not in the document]
calmacx: https://dev.azure.com/UniversityOfNottingham/DRS/_backlogs/backlog/CO-CONNECT/Epics/?workitem=35351

api/mapping/services_rules.py
calmacx: https://dev.azure.com/UniversityOfNottingham/DRS/_backlogs/backlog/CO-CONNECT/Epics/?workitem=35352

PR #117, opened 2021-05-13
35228/refactor nlp

simon-tarr:
This PR contains refactored NLP code. At present, the code isn't wrapped in a Celery task, so I would avoid running this on a large list of values. In addition, you must provide descriptions via the admin interface for ScanReportValues, or via the form in the web mapping tool for ScanReportFields. As an aside, it might be worth creating a DevOps task to allow _value_ descriptions to be added/updated from the web tool interface.

The functionality operates at two levels: the Field and Value level.

If you set ScanReportField to `pass_from_source`, this means that we only send the _field description_ to the NLP service. In other words, we don't want to map individual ScanReportValue's for this field. Fields you'd typically set as `pass_from_source` are height, age, weight etc (i.e. continuous variables).

If ScanReportField is _not_ set to `pass_from_source` then the code will process _all_ values associated with that field. Once the code has finished running, conceptIDs from NLP will be displayed in blue. Black text is reserved for those conceptIDs that a user has manually inputted.

You won't get any return from NLP if (1) It simply can't find a term for your text string (2) You've forgotten to add a field/value description, so you end up sending a blank field to NLP.

NOTE - At the moment, the redirect forwards the user to the _values_ page associated with the field. This is so that it's easy to see when you come back to the task what field you ran it on (if I just returned them to the scanreportfield page, there's no way to tell)


api/mapping/services_nlp.py
```diff
@@ -1,19 +1,305 @@
...
+    for url in post_response_url:
 
-def nlp_single_string(pk, dict_string):
+        req = requests.get(url, headers=headers)
+        job = req.json()
+
+        while job["status"] != "succeeded":
+            req = requests.get(url, headers=headers)
+            job = req.json()
+            print("Waiting...")
```
py5gol, line 31: Replace with logging, please.

api/mapping/services_nlp.py
```diff
@@ -1,19 +1,305 @@
...
+            time.sleep(3)
+        else:
+            get_response.append(job["results"])
+            print("Done!")
```
py5gol, line 35: Replace with logging, please.

api/mapping/services_nlp.py
```diff
@@ -1,19 +1,305 @@
...
+                                    dict_entry["id"],
+                                    entity["text"],
+                                    entity["category"],
+                                    entity["confidenceScore"],
+                                    link["dataSource"],
+                                    link["id"],
+                                ]
+                            )
+
+    print('CODES >>>', codes)
```
py5gol, line 69: Replace with logging, please.

api/mapping/services_nlp.py
```diff
@@ -1,19 +1,305 @@
...
+        )
+        item.append(x[1].concept_id)
+        codes_dict.append(dict(zip(keys, item)))
+
+    return codes_dict
+
+
+def start_nlp(search_term):
+
+    print(">>>>> Running NLP in services_nlp.py for", search_term)
```
py5gol, line 100: Replace with logging, please.

api/mapping/services_nlp.py
```diff
@@ -1,19 +1,305 @@
...
+    field = ScanReportField.objects.get(pk=search_term)
```
py5gol, line 101: What is search_term? Is it the scanreportfield id?

api/mapping/services_nlp.py
```diff
@@ -1,19 +1,305 @@
...
+    scan_report_id = field.scan_report_table.scan_report.id
+
+    # Define NLP things
+    url = "https://ccnett2.cognitiveservices.azure.com/text/analytics/v3.1-preview.3/entities/health/jobs?stringIndexType=TextElements_v8"
```
py5gol, line 105: Can we extract this to the env file?

api/mapping/services_nlp.py
```diff
@@ -1,19 +1,305 @@
...
+
+    # Create empty list to later hold job URLs
+    post_response_url = []
 
+    # Checks to see if the field is 'pass_from_source'
+    # If True, we pass field-level data i.e. a single string (field description)
+    # If False, we pass all values associated with that field
+    if field.pass_from_source:
+
+        print(">>> Working at field level...")
```
py5gol, line 119: Replace with logging, please.

api/mapping/services_nlp.py
```diff
@@ -1,19 +1,305 @@
...
+                nlp_entity=item["nlp_entity"],
+                nlp_entity_type=item["nlp_entity_type"],
+                nlp_confidence=item["nlp_confidence"],
+                nlp_vocabulary=item["nlp_vocab"],
+                nlp_concept_code=item["nlp_code"],
+                concept=concept,
+                content_object=scan_report_field,
+            )
+
+    else:
```
py5gol, line 173: These two if conditions are long, and could got into functions to improve readability.

api/mapping/services_nlp.py
```diff
@@ -1,19 +1,305 @@
...
+            scan_report__id=scan_report_id)
+        neg_assertions = assertions.values_list("negative_assertion")
+
+        # Grab values associated with the ScanReportField
+        # Remove values in the negative assertions list
+        scan_report_values = ScanReportValue.objects.filter(
+            scan_report_field=search_term
+        ).filter(~Q(value__in=neg_assertions))
+
+        # Create list of items to be sent to the NLP service
```
py5gol, line 187: This section could/should be a functon.

api/mapping/services_nlp.py
```diff
@@ -1,19 +1,305 @@
...
+
+        # Mini function to see if all conceptIDs are the same
+        def all_same(items):
+            return all(x == items[0] for x in items)
+
+        # Check each item in values and see whether NLP got a result
+        # If NLP finds something, save the result to ScanReportConcept
+        # If all conceptIDs across vocabs are the same, save only SNOMED
+        # Else save each conceptID to ScanReportConcept
+        for value in scan_report_values:
```
py5gol, line 249: This could/should be in another function.

api/mapping/services_nlp.py
```diff
@@ -1,19 +1,305 @@
...
+                            nlp_entity=item["nlp_entity"],
+                            nlp_entity_type=item["nlp_entity_type"],
+                            nlp_confidence=item["nlp_confidence"],
+                            nlp_vocabulary=item["nlp_vocab"],
+                            nlp_concept_code=item["nlp_code"],
+                            concept=concept,
+                            content_object=scan_report_value,
+                        )
+
+    return True
```
py5gol, line 299: Why does it return True?

api/mapping/services_nlp.py
```diff
@@ -1,19 +1,305 @@
...
+
+
+def nlp_single_string(pk, dict_string):
```
py5gol, line 302: This function name could reflect what happens in it. Example: get_concept_id_list_from_nlp_string?

api/mapping/services_nlp.py
```diff
@@ -67,12 +353,12 @@ def nlp_single_string(pk, dict_string):
 
     return True
```
py5gol, line 354: Why does it return True?

api/mapping/services_nlp.py
```diff
@@ -67,12 +353,12 @@ def nlp_single_string(pk, dict_string):
...
 def get_json_from_nlpmodel(json):
-    
     """
     A small function to process the JSON string saved in NLPModel
     """
-    
+
     # Define which codes we want to keep. Add more here as required.
     codes = []
     keep = ["ICD9", "ICD10", "RXNORM", "SNOMEDCT_US"]
```
py5gol, line 364: I think this list is defined multiple times. Can you extract it to a common place?

api/mapping/services_nlp.py
```diff
@@ -114,8 +400,7 @@ def get_json_from_nlpmodel(json):
     full_results = pd.concat(results, ignore_index=True)
```
py5gol, line 400: Pandas??

api/mapping/templates/mapping/scanreportvalue_list.html
```diff
@@ -2,64 +2,84 @@
...
+    </ol>
+</nav>
 
-    <h2>Values</h2>
-
-    <div>
-        <table class="table" style="float:left;">
-            <thead>
+<h2>Values</h2>
+{{scan_report_concept}}
```
py5gol, line 21: Do we need it here?

py5gol (changes requested):

api/mapping/services_nlp.py
simon-tarr: @py5gol I've been meaning to speak to you about this. I think I mentioned it previously but I was wondering if we ought to keep our list of vocabs in a mini-model so they are available across the whole site (if needed). Might be a bit overkill, so let me know what you think.

api/mapping/services_nlp.py
simon-tarr: Yes. Do you want me to change this to something else @py5gol?

api/mapping/services_nlp.py
simon-tarr: Sure, but why? We only use it once in the whole application.

api/mapping/services_nlp.py
simon-tarr: Does the function have to return anything? It's not actually returning any data back to the view - it's job is to just save data to ScanReportConcept. Once it finishes opn line 297 there is nothing left to return back to views.py.

I set `return True` as I wasn't sure what else I should return.

api/mapping/services_nlp.py
simon-tarr: I've deleted this function because we no longer need it now NLP is integrated at the field/value level.

api/mapping/templates/mapping/scanreportvalue_list.html
simon-tarr: Nope, deleted! Must've been a hangover from something I was testing.

api/mapping/services_nlp.py
simon-tarr: Removed.

api/mapping/services_nlp.py
simon-tarr: As an aside, now that the old functions at the bottom of services_nlp.py have been removed, this list does only apear once in the entire NLP code. A model may therefore be overkill. Let me know what you think @py5gol

py5gol (approved):

PR #122, opened 2021-05-19
35346/field concept i ds

vpnu:
PR for adding multiple concept IDs on the field level

### DIFF
--- a/api/mapping/models.py
+++ b/api/mapping/models.py
@@ -468,6 +468,7 @@ class ScanReportField(BaseModel):
     concepts = GenericRelation(
         ScanReportConcept,
     )
+    
     def __str__(self):
         return str(self.id)
 
--- a/api/mapping/models.py
+++ b/api/mapping/models.py
@@ -460,6 +460,9 @@ class ScanReportField(BaseModel):
 
     concept_id = models.IntegerField(
         default=-1,
+        blank=True,
+        null=True,
+        # This field is not used anymore
     )
 
     concepts = GenericRelation(
--- a/api/mapping/models.py
+++ b/api/mapping/models.py
@@ -460,10 +460,11 @@ class ScanReportField(BaseModel):
 
     concept_id = models.IntegerField(
         default=-1,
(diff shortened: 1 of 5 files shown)


api/mapping/views.py
```diff
@@ -148,11 +149,12 @@ def get_success_url(self):
         )
 
 @method_decorator(login_required, name="dispatch")
-class ScanReportFieldListView(ModelFormSetView):
+class ScanReportFieldListView(ListView):
     model = ScanReportField
-    fields = ["is_patient_id", "date_type", "concept_id"]
-    fields = ["is_patient_id", "is_birth_date", "is_date_event", "concept_id"]
-    # exclude = []
+    # fields = ["is_patient_id", "date_type", "concept_id"]
```
py5gol, line 154: @Vas, you can delete lines 154-156.

api/mapping/views.py
```diff
@@ -1504,3 +1506,48 @@ def delete_scan_report_value_concept(request):
...
+    scan_report_concept = ScanReportConcept.objects.get(pk=scan_report_concept_id)
+
+    concept_id = scan_report_concept.concept.concept_id
+    concept_name = scan_report_concept.concept.concept_name
+
+    scan_report_concept.delete()
+
+    messages.success(request, "Concept {} - {} removed successfully.".format(concept_id, concept_name))
+
+    return redirect("/fields/?search={}".format(scan_report_table_id))
```
py5gol, line 1553: Missing a new line at the end of the file.

api/mapping/models.py
```diff
@@ -460,14 +460,11 @@ class ScanReportField(BaseModel):
 
     concept_id = models.IntegerField(
         default=-1,
-        null=True,
```
py5gol, line 463: Do we need to remove null and blank? If not, I would leave them here. And, I would add a note saying this field is not used anymore.

py5gol (changes requested):

py5gol (approved):

PR #132, opened 2021-05-28
Shahzad mapping pipeline api

shahzadmumtaz22:


spco:
Merged into devel after the move to simple appservice deployment thru Actions.

PR #137, opened 2021-06-03
[IMPORTANT] Fix postgres db image tag, to keep consistency 

calmacx:
I found that the postgres running on the ccnettestdb or on the VM might be different from what starts running if you spin up the docker-compose locally where the postgres docker images is just taking `postgres:latest`


calmacx:
irrelevant now we dont use the db via docker-compose

PR #144, opened 2021-06-08
Vas azure function

vpnu:
**PR for Azure Function ProcessQueue**

- Includes the changes in settings.py for uploading the Scan Report to blob storage using the Django-storage[azure] package

- Shahzad's API branch was merged into this, plus some changes he provided to allow for JSON Array as input on the API side.


simon-tarr:
Hey @spco, @vpnu,

Not necessarily a discussion for this PR but we must (as a team) decide how we want to track, deploy and maintain Azure functions. At the moment we have:

1) No formal way of tracking Azure functions in Git. Are we putting them all into a folder called 'azure_functions' in the project's root dir? Or are we having individual dirs for each function in root? e.g. a dir for ScanReports, another for NLP.

2) We haven't established a way of creating key vaults for sharing API keys and the like in deployed azure functions. At present, secrets are kept in local.settings.json and this file acts as a .env file for local testing of Azure functions. We need to move this to key vault for these functions to work when deployed (https://medium.com/weareservian/accessing-azure-key-vault-from-python-functions-44d548b49b37)

At the moment, neither mine or Vas' functions would actually work when deployed (and we definitely shouldn't be tracking local.settings.json in Git!)

spco:
Hi @simon-tarr - all 3 of those are very valid questions, and hopefully the answer to at least (1) and (2) will become clearer during the move to AppService (they may be pretty much dictated to us, not sure right now)

PR #160, opened 2021-06-24
Generated new secret key and masked it with os.environ

simon-tarr:
PR to mask secret key. 
Key must now be kept in .env


spco:
CI failed because `SECRET_KEY` was not set in environment. Will set that now.

PR #161, opened 2021-06-24
Update/submodule usage

calmacx:
We shouldnt be needing to rely on the co-connect-tools submodule anymore, this adds complication.

OmopDetails is a class that has been long redundant, so it is removed in services.py 

Since we still use `coconnect.tools.make_dag()` function, we install `co-connect-tools` via pip instead

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,10 +37,6 @@ RUN pip install -r /api/requirements.txt --no-cache-dir
 USER root
 RUN apt-get install graphviz -y
 
-COPY --chown=django:django co-connect-tools/ /coconnect/
-
 USER django
-
-RUN pip install -e /coconnect/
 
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -37,10 +37,6 @@ RUN pip install -r /api/requirements.txt --no-cache-dir
 USER root
 RUN apt-get install graphviz -y
```
spco, line 38: Could you also move this install up to the main `apt-get` command above please?

spco:
In general I think this is certainly an improvement on the current setup 👍 . We can discuss at a later date whether we really need to install the whole of `co-connect-tools` just to access a single function, but that can wait.

Dockerfile
calmacx: good point. done

PR #162, opened 2021-06-24
Update ci.yml

spco:


calmacx:
CI is failing probably because of bad characters, quotations should do the trick:
```
echo "SECRET_KEY=${{ secrets.SECRET_KEY }}" >> .env
```

PR #163, opened 2021-06-24
Dev/add spinners

calmacx:
* spinner moved from raw html in javascript to a div container that can be toggled
* refresh rules and download json now moved to ajax post request
* all three buttons now display the spinner while they are waiting for a response from posting data 
* download json has some complex ajax code (stolen from stackoverflow) for downloading content response from a post method, that should in theory work for multiple browsers


spco:
@calmacx could you update this to fix the merge conflict? I don't understand the ajax at all, but I'm happy to run with what you have. Is this OK to merge by you (once merge conflict is fixed)? I will do a squash as this doesn't need 23 commits in the history :)

calmacx:
> @calmacx could you update this to fix the merge conflict? I don't understand the ajax at all, but I'm happy to run with what you have. Is this OK to merge by you (once merge conflict is fixed)? I will do a squash as this doesn't need 23 commits in the history :)

done. 90% of the ajax is a copy and paste from stackoverflow, as I linked in the source code, worked for me on multiple browsers :) 

If you're interested... that complex bit of code is because the post used to just return a json response, that would download automatically. But now since the post is made via ajax, so a spinner can be displayed saying it's waiting for a response, there needs to be a way of rendering the returned content and converting it to an auto download again - turns out that's a complex problem in ajax

PR #165, opened 2021-06-24
More concise code for document creation

calmacx:


spco:
This is fine, but a bit strangely structured still because the `None`-ness of `item.scan_report_field.description_column` and `item.value_description` are entirely independent. Would it be clearer to instead say

```
field_text = item.scan_report_field.description_column if item.scan_report_field.description_column is not None else item.scan_report_field.name
value_text = item.value_description if item.value_description is not None else item.value
document = {
  "documents":[
    {
      "language": "en", "id": str(item.id)+'_value',
      "text": field_text.replace("_", " ")+', '+value_text.replace("_", " ")
     }
  ]
}
```
?

simon-tarr:
Yep leave it with me @spco and @calmacx. Thanks both for taking a look and improving this block of code - I knew it was a bit verbose when I wrote it, so it's nice to see something leaner.

simon-tarr:
Hi @spco, I've integrated your suggested code into the app. Note that when evaluating `pass_from_source=True`  I check for a literal blank string (i.e. ""). For example on Line 117:
```
field_text = field.description_column if field.description_column is not "" else field.name
```
Whereas if pass_from_source=False (i.e. running at the value level) then I evaluate on `None`. For example on Line 159:
```
field_text = item.scan_report_field.description_column if item.scan_report_field.description_column is not None else item.scan_report_field.name

```

This is due to how Vas' Azure Function reads in data from the scan report. For fields, there is a field_description column; this could be blank or contain a descriptive string of text. At the moment, Vas' code casts field_description to string (`str(field_description)`) which causes literal blank strings when processing each row. There are no fields in the scan report for value descriptions and there explictly set as `None` in her code.

We may want to decide at a later date whether we want to evaluate on "" _or_ `None` (rather than using both as in the current implementation)

spco:
Thanks @simon-tarr - can you raise an Issue here on GH to capture that question please?

Otherwise, are you ok for me to merge this PR? Again, this is a one-commit PR :)

simon-tarr:
Happy for you to merge this PR into devel @spco. I've raised an Issue as requested.

PR #167, opened 2021-06-28
Fix bug when adding concepts not in m_allowed_tables

calmacx:
* Refresh Rules fails because of drug_exposure concepts that have been added. Now it failed to generate rules for already existing ones
image

* Stop non-allowed concepts from being added in the first place
image

### DIFF
--- a/api/mapping/services_rules.py
+++ b/api/mapping/services_rules.py
@@ -59,12 +59,18 @@ def get_omop_field(destination_field,
     Returns:
       - OmopField : the destination field object
     """
+
     #if we haven't specified the table name
     if destination_table == None:
         #look up the field from the "allowed_tables"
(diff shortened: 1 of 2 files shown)


api/mapping/services_rules.py
```diff
@@ -59,12 +59,18 @@ def get_omop_field(destination_field,
...
       - OmopField : the destination field object
     """
+
     #if we haven't specified the table name
     if destination_table == None:
         #look up the field from the "allowed_tables"
         omop_field = OmopField.objects\
-                               .filter(table__table__in=m_allowed_tables)\
-                               .get(field=destination_field)
+                              .filter(field=destination_field)
```
calmacx, line 67: this was previously crashing when None was retrieved due to the table not being in m_allowed_tables

PR #168, opened 2021-06-28
Feature/add drug exposure table

calmacx:
depends on #167 

image

### DIFF
--- a/api/mapping/services_rules.py
+++ b/api/mapping/services_rules.py
@@ -15,14 +15,15 @@ class NonStandardConceptMapsToSelf(Exception):
 
            
 #allowed tables
-m_allowed_tables = ['person','measurement','condition_occurrence','observation']
+m_allowed_tables = ['person','measurement','condition_occurrence','observation','drug_exposure']
 
 #look up of date-events in all the allowed (destination) tables
 m_date_field_mapper = {
     'person': ['birth_datetime'],
     'condition_occurrence': ['condition_start_datetime','condition_end_datetime'],
     'measurement':['measurement_datetime'],
-    'observation':['observation_datetime']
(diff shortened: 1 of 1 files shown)


calmacx:
> @calmacx what stage is this PR at? Thanks.

@spco i actually spoke with Erum about this yesterday and she said for the drug_exposure table they dont need anything in addition, meaning that this PR is good to go now actually. 

They only need to 6 rules shown in the screenshot, nothing else (at current).

spco:
Ok, so this now just contains the changes in `services_rules.py` - to clarify, that is all that's currently needed?

calmacx:
hi @spco indeed this is all that is needed to turn on this particular table in the mapping tool.

What don't you think matches up with the ticket? In the ticket they say the only fields they need are the person_id, concept_id and start date. By turning this table on, they'll also get the source_value and source_concept_id (which IMO should be included) and the end date (which can be removed if really necessary)

Or am I missing something? cheers

spco:
This doesn't explicitly mention `drug_concept_id` - so I'm guessing that is exposed through the `drug_exposure` table? I was just thinking that your force-push removed quite a lot of code so wanted to check.

calmacx:
> This doesn't explicitly mention `drug_concept_id` - so I'm guessing that is exposed through the `drug_exposure` table? I was just thinking that your force-push removed quite a lot of code so wanted to check.

That's right, I have a dump of the Omop table/fields we are using (v5.3.1) here: https://github.com/CO-CONNECT/co-connect-tools/blob/master/coconnect/data/cdm/OMOP_CDM_ONLINE_LATEST.csv#L32

Yeh no worries, the force was from a rebase and it does look odd that this is all that is needed :)

PR #169, opened 2021-06-28
Fix/process queue post

vpnu:
Posting values per sheet(table) and fixing bug where field and value names were empty. Plus fixing the blank field description from "empty" to match the actual value from the scan report file.


ProcessQueue/__init__.py
```diff
@@ -215,19 +233,20 @@ def main(msg: func.QueueMessage):
         # Create a dictionary with field names and field ids 
         # as key value pairs
         # e.g ("Field Name":<Field ID>)
-        namexids=dict(zip(field_names,field_ids))
-
+        namexids=dict(zip(field_ids,field_names))
+        
         # For sheets past the first two in the Scan Report
         # i.e. all 'data' sheets that are not Field Overview and Table Overview
         data=[]
```
spco, line 240: Please remove as this is unused.

ProcessQueue/__init__.py
```diff
@@ -215,19 +233,20 @@ def main(msg: func.QueueMessage):
         # Create a dictionary with field names and field ids 
         # as key value pairs
         # e.g ("Field Name":<Field ID>)
-        namexids=dict(zip(field_names,field_ids))
-
+        namexids=dict(zip(field_ids,field_names))
```
spco, line 236: Has the order of this changed because that's what `get_field_id()` expects? If so, could you rename this to `id_x_names` to better reflect its structure? Thanks.

ProcessQueue/__init__.py
```diff
@@ -248,8 +267,14 @@ def main(msg: func.QueueMessage):
                 
                 # If we are not on the first row:
                 if name!=value:
-                    print(result)
-                    print("Frequency",frequency)
+                    #Get the field_id
+                    #If more than one field with the same name, pick the one in the current sheet
+                    field_id=get_field_id(namexids,field=name)
+                    if len(field_id)>1:
+                        field_id=field_id[idxsheet-2]
```
spco, line 274: Will this logic work correctly if some but not all of the sheets contain the same field name? So if we have 3 tables and table 1 and table 3 contain `my_field_name` then `get_field_name` will return 1 list of length 2, which wouldn't then match the `idxsheet-2` index used.

spco (commented):
I will likely also comment further on the the lines 40-60 but just completing this review now for you to see it. Let me know if I've misunderstood anything.

ProcessQueue/__init__.py
vpnu: This is used on line 272 to match the values to the fields

ProcessQueue/__init__.py
vpnu: ah yes you're right I'll try to come up with a better solution

ProcessQueue/__init__.py
spco: Sorry, I was referring to `data` which is reset within the loop below.

ProcessQueue/__init__.py
```diff
@@ -32,27 +32,42 @@ def process_scan_report_sheet_table(sheet):
     # Get max number of columns in the sheet
     max_column=sheet.max_column
     # Set a column index (openpyxl index at 1 instead of 0)
-    col_idx=1
     
     for row_idx,row_cell in enumerate(sheet.iter_rows(min_row=1,max_col=max_column),start = 1):
   
     # Works through pairs of value/frequency columns
         for column_idx,cell in enumerate(row_cell):
```
spco, line 39: What is the pseudocode for this process? I feel like it could be tidied up somewhat. Is it this?
```
Loop over all columns. Skip the frequency columns, only process the value columns.
  For each value column, loop over all rows - if either the value column or its partnered frequency column is non-empty, then process this.
    Append the (column-name,value,frequency) tuple to results 
```

spco: In fact, I think we should be looping over columns first, and then rows (because columns have different lengths).
```
for col_idx, column in enumerate(sheet.iter_cols(min_row=1, max_row=sheet.max_row), start=1):
  if col_idx % 2  == 0:
    continue
  for row_idx, cell in enumerate(column):
    if cell.value == "" and sheet.cell(row=row_idx,column=column_idx+2).value == "":
      continue
    results.append(  (sheet.cell(row=1,column=column_idx+1).value, cell.value, sheet.cell(row=row_idx,column=column_idx+2).value)
```

spco: (I have not tested the above)

ProcessQueue/__init__.py
vpnu: ah yeah you mean I'm emptying it again as I enter the for loop on line 243

ProcessQueue/__init__.py
vpnu:
```
for every row in a sheet:
  for every column in the row:
      If cell.value is not empty then:
         make sure to take pairs of value/frequency columns (%2=0)
          if both value and frequency are empty, continue (this can be removed I think, if cell.value takes care of this)
           append the (column-name,value,frequency) to results 
      Else, 
             if cell.value is empty but there is a frequency
               append column-name,value==empty,frequency) to results 

```

spco: I think we should certainly move the `(%2=0)` check as high as possible because we want to ignore the frequency columns immediately, and just process the value columns (with the associated freq of course). And combine the check for a value in the cell with the check for the value in the associated frequency - we only do something if either is non-empty, otherwise we ignore.

(What happens if value is nonempty but frequency is empty, or can that never happen?)

vpnu: if there is a value but no frequency I set it to be 0 in line 273-276

vpnu: if we want it to be empty instead of zero we can allow it to be blank like we have the field, value names

spco: Ok, I'm happy to set it to zero if it's empty. So if either the value or the freq is non-empty then we should add this value-freq pair to `results`, and we only `continue` if both are empty - is that correct?

I think it's conceptually clearer to write it as 
```
# Iterate over all columns in this sheet
for col_idx, column in enumerate(sheet.iter_cols(min_row=1, max_row=sheet.max_row), start=1):
  # Only process the value columns, and skip the frequency columns
  if col_idx % 2  == 0:
    continue
  # Loop over all rows in this column
  for row_idx, cell in enumerate(column):
    # Skip this value if it and its frequency are empty
    if cell.value == "" and sheet.cell(row=row_idx,column=column_idx+2).value == "":
      continue
    # Otherwise, one of the value-freq pair must be non-empty, and so we add the value-freq pair to results
    results.append((sheet.cell(row=1,column=column_idx+1).value, cell.value, sheet.cell(row=row_idx,column=column_idx+2).value))
```

ProcessQueue/__init__.py
```diff
@@ -30,76 +31,119 @@ def process_scan_report_sheet_table(sheet):
...
     return results
 
 
+def get_field_id(dict, field):
+    """
+    Input: dictionary with field_id:field_value pairs
+    Input: field name
+    Returns: field id
+    """
+    for id, name in dict.items():
```
spco, line 84: The more I think about this the more I think it can be simplified. As it stands, we create `id_x_names` as a dictionary of `id:name`. We then use `get_field_id()` to do a reverse-lookup: given a name, return the id associated to it. And this just returns the first matching name, because we assume that the name will be unique. This is the only place `id_x_names` is used.

If that's the case, then we might as well create a reverse dictionary `name:id` and just do a lookup. If need be, we can add in a guard to flag any time when it tries to add a second value to the same `name`. Then we can get rid of this whole `get_field_id()` function.

ProcessQueue/__init__.py
```diff
@@ -111,161 +155,189 @@ def main(msg: func.QueueMessage):
...
+                print("FIELD SAVE STATUS >>>", response.status_code)
+                # Load result from the response,
+                # Save generated field ids, and the corresponding name
+                response = json.loads(response.content.decode("utf-8"))
+                for element in range(len(response)):
+                    field_ids.append(str(response[element].get("id", None)))
+                    field_names.append(str(response[element].get("name", None)))
+                # Fix for empty row (empty row was being returned as [] element)
+                # Need to remove for dict(zip()) to work
+                for name in field_names:
```
spco, line 269: In this `for` loop and the next, can we be _absolutely_ certain that all `"["` or `None` in `field_ids` will match up exactly with all `"None"` in `field_ids`? If not, we could end up misaligning them. Is there a safer way to do this? (perhaps iterate over both at the same time with a `zip` and only delete when both match the criteria - I just think this would be safer and avoid some difficult-to-notice bug with misaligned lists).

spco (commented):
This is really great, I'm so pleased it's now working. Please take a look at the comments I've made as I think they will make it more robust - and we can discuss anything that's not clear or you disagree with!

ProcessQueue/__init__.py
```diff
@@ -30,76 +31,119 @@ def process_scan_report_sheet_table(sheet):
...
-                # This is required b/c value/frequency col pairs differ
-                # in the number of rows
-                  if sheet.cell(row=row_idx,column=column_idx+1).value == "" and sheet.cell(row=row_idx,column=column_idx+2).value == "":
-                    continue
-                # Append to Results as (Field Name,Value,Frequency)
-                  results.append(
-                      (sheet.cell(row=1,column=column_idx+1).value,sheet.cell(row=row_idx,column=column_idx+1).value, sheet.cell(row=row_idx,column=column_idx+2).value)
-                  )
-    col_idx=col_idx+1
+                    # As we move down rows, checks that there's data there
```
spco, line 46: Please move `if (column_idx) % 2 == 0:` out to before `if cell.value:`: in both branches  of the `if cell.value / else` we only do something if `(column_idx) % 2 == 0`, so it's clearer to check for it in only one place.

ProcessQueue/__init__.py
vpnu: Yeah that makes sense.This used to be name:id but I switched it around because names weren't always unique. So since it's no longer an issue I'll change it back to that and remove the get_field_id()

ProcessQueue/__init__.py
spco: In fact, I think this whole thing can still be simplified.
```
if sheet.cell(row=row_idx, column=column_idx + 1).value or sheet.cell(row=row_idx, column=column_idx + 2).value:
  results.append(
                        (
                            sheet.cell(row=1, column=column_idx + 1).value,
                            sheet.cell(row=row_idx, column=column_idx + 1).value,
                            sheet.cell(row=row_idx, column=column_idx + 2).value,
                        )
                    )
```
or am I missing something? This can be made even clearer if we set `start=1` in the `enumerate(row_cell)` to 
```
if sheet.cell(row=row_idx, column=column_idx).value or sheet.cell(row=row_idx, column=column_idx + 1).value:
  results.append(
                        (
                            sheet.cell(row=1, column=column_idx).value,
                            sheet.cell(row=row_idx, column=column_idx).value,
                            sheet.cell(row=row_idx, column=column_idx + 1).value,
                        )
                    )
```

ProcessQueue/__init__.py
vpnu: Yeah sure I can try that, I don't remember what exactly was causing this or why I fixed it like this so I'll have a look and see if this is still an issue with the changed code in the first place

spco: Sure - if we don't get those same "[" or "None" outputs then we can get rid - otherwise, let's go with the safer implementation.

ProcessQueue/__init__.py
vpnu: I think I tried this before:
if sheet.cell(row=row_idx, column=column_idx).value or sheet.cell(row=row_idx, column=column_idx + 1).value:
 and it was giving me bad request at the values, but that was back when we weren't allowing blank value names I think so that might be why. I can try to change it around see if we get the same results

PR #171, opened 2021-06-30
Added messages to say NLP running (rather than a redirect)

simon-tarr:
Small PR to add a message banner for NLP run (instead of a redirect to values).


spco (approved):
Looks good to me! Please test once deployed.

PR #188, opened 2021-07-06
37688/remove xlsx2csv

simon-tarr:


api/mapping/forms.py
```diff
@@ -34,35 +35,65 @@ class ScanReportForm(forms.Form):
...
+
+            # Load in the Excel sheet, grab the first workbook
+            file_in_memory = self.cleaned_data['scan_report_file'].read()
+            wb = openpyxl.load_workbook(filename=BytesIO(file_in_memory), data_only = True)
+            ws=wb.worksheets[0]
+
+            # Grab the scan report columns from the first worksheet
+            # Define what the column headings should be
+            source_headers = []
+            for values in ws[1]: source_headers.append(values.value)
```
spco, line 47: Please split onto indented line

api/mapping/forms.py
```diff
@@ -34,35 +35,65 @@ class ScanReportForm(forms.Form):
...
+            expected_headers=['Table', 'Field', 'Description', 'Type', 'Max length', 'N rows', 'N rows checked', 'Fraction empty', 'N unique values', 'Fraction unique', 'Flag', 'Classification']
+            
+            # Check if source headers match the expected headers
+            if set(source_headers) == set(expected_headers):
```
spco, line 51: `set()` doesn't preserve the ordering, so please change this to be a direct comparison of the lists - otherwise we risk ingesting files with all the correct columns but in the wrong order.

api/mapping/forms.py
```diff
@@ -34,35 +35,65 @@ class ScanReportForm(forms.Form):
...
+
+                # Removes the column name (here, 'Classification') from the list of Flag values
+                classification_column_data.pop(0)
+
+                # Define what flags and classifications we allow in respective columns
+                allowed_flags = ['PATIENTID', 'DATE', 'IGNORE', 'PASS_SOURCE']
+                allowed_classifications = ['SNOMED', 'RXNORM', 'ICD9', 'ICD10']
+
+                # Test whether the values in the flag column are in our list of allowed flags
+                if all(x in allowed_flags for x in list(filter(None, flag_column_data))):
```
spco, line 82: Turn this into an `if not ...` and then you can remove the `pass` and `else` statements.

api/mapping/forms.py
```diff
@@ -34,35 +35,65 @@ class ScanReportForm(forms.Form):
...
+                    pass
+                else:
+                    raise (forms.ValidationError("Check 'Flag' column values. Valid options are " + ', '.join(allowed_flags)))
+
+                # Test whether the values in the classificcation column are in our list of allowed classifications
+                if all(x in allowed_classifications for x in list(filter(None, classification_column_data))):
```
spco, line 88: ditto

api/mapping/forms.py
spco: also, please rename `x` to `flag` or similar (I know, my example code had `x` in!)

api/mapping/forms.py
```diff
@@ -34,35 +35,65 @@ class ScanReportForm(forms.Form):
...
-                            pass
-                        else:
-                            raise (forms.ValidationError( "Check Flag column values. Valid options are: {} or blank".format(list(FLAG_CHOICES.values()))))
-                        
-                        if (classification_column in VOCABULARY_CHOICES.values()) or (classification_column==''):
-                            pass
-                        else:
-                            raise (forms.ValidationError( "Check Classification column values. Valid options are:{} or blank".format(list(VOCABULARY_CHOICES.values()))))
-                    return self.cleaned_data['scan_report_file']
+
```
spco, line 38: swap this `if` for `if not` and then the `ValidationError` near the bottom can be dragged up here - reduces the indenting and is a little clearer in showing it will abort most of the function.

api/mapping/forms.py
spco: And see the `if not` comment as above - drag the related `ValidationError` up to here.

spco: "as above" in this instance is in a comment further down here.

simon-tarr:
@spco code is ready for you again!

api/mapping/forms.py
spco: This one still has the `else` at the bottom - change it to `if not` and then put the `ValidationError` in here.

PR #190, opened 2021-07-07
Deleted old Usagi services file

simon-tarr:


spco:
Now this is the kind of PR I like 😆

PR #212, opened 2021-07-20
Fix/bug/spinning internal error message

calmacx:
### The Problem

* when clicking on refresh rules, the ajax returns as Internal Server Error message

image

### The Cause

* The data team had managed to add a concept from a table (Specimen) that is not allowed... 
image

* This must have been added BEFORE we had included extra validation for adding concepts from tables that we don't support

* Additional problem because the code was validating by looking up `specimen_source_concept_id` as it does with all other domains, and this one does not exist in OMOP
image

### DIFF
--- a/api/mapping/services_rules.py
+++ b/api/mapping/services_rules.py
@@ -71,6 +71,8 @@ def get_omop_field(destination_field,
 
         if len(omop_field)>1:
             return omop_field.filter(table__table__in=m_allowed_tables)[0]
+        elif len(omop_field) == 0:
+            return None
         else:
             return omop_field[0]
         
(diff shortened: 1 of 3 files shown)


spco:
@calmacx is this ready to review? What is the behaviour now - it shows an error message when the page with the incorrect concept loads, and shows it as a blank in the table?

calmacx:
@spco it is indeed ready for review. 

* Now if the data team try to add a specimen concept, it will not be allowed, nor will cause a crash, and will show the error message in the screenshot
* When generating/refreshing the rules, this rule for specimen should get skipped and wont display in the table. However I need to verify on the test server, because locally im having the issue of the page timing out again, now that I refreshed the rules:
image

Since my local db is pointing to dev-db, you can see that it has fixed the rules on https://ccom-dev.azurewebsites.net/scanreports/101/mapping_rules/

calmacx:
side note: Actually the refresh rules works on https://ccom-dev.azurewebsites.net/scanreports/101/mapping_rules/ because the bad concept was deleted

PR #214, opened 2021-07-20
Code cleanup

vpnu:

PR #215, opened 2021-07-23
Merge dev to test 230721

spco:
Contains:

- Use whitenoise and gunicorn (#206)
- Fix bundling of static files. (#204 , #213)
- Update ProcessQueue to new package dependency and structure including data dictionaries (#208)
- Fixed JSON output bug (#211)
- Lots of tidying of unused code and imports, and move `make_dag()` into the code to remove dependence on co-connect-tools. (#214 , #209)
- Add queries to ScanReportValueSerializer (#210)
- Fix error caused by invalid concept (#212)


spco:
Note that this required the creation of `scan-reports` container in test storage account. I have created that there, and another `scan-reports` in the _prod_ system as well ahead of merging there.

PR #220, opened 2021-07-28
Fix/mapping rules load time

calmacx:


spco:
I like this in general. Do you have any feel for how this performs in comparison?

calmacx:
When I tested it locally it didn't seem to make a massive difference. I jotted the numbers down at it went from 17 seconds to 15 seconds to load (that was using Chrome dev tools and could well have been consistent with each other) 

Testing has the same problem we discussed before, no idea if any speed ups are significant or not, because on dev-db we don't have many examples of ScanReports that have been mapped with lots of concepts and are taking >30 seconds to load.

I'm "hacking" the `get_context_data` function, so it could be that the `super().get_context_data` that builds the original `context['object_list']` is slow and making many queries.  This is what I was guessing, got no proof of this yet and depends what @simon-tarr has found from his tests/profiling.

I don't still quite understand the order that Django executes these various methods when preparing data for the templating. Nor have I timed this new loop I implemented - to see if it really is the GenericRelation via `content_object` that is the thing that is super slow for the mapping-rules page.

calmacx:
Closing this as is superseded by #235

PR #222, opened 2021-07-30
Convert concept codes to conceptIDs on Scan Report upload

simon-tarr:
This PR makes the following changes:

1. Creates a new folder in root called `shared_code`. Within `shared_code` I have created `omop_helpers.py` which contains OMOP-related functions (e.g. `get_concept_from_concept_code()` etc.) This allows Azure Functions to share methods across different functions (e.g. we can use methods in `omop_helpers.py` within NLPQueue or ProcessQueue with an import at the top of `__init__.py`. This means that we only have to maintain these methods from a single location. Note that this PR--while focussing on concept code -> conceptID lookup--does move functions used by NLPQueue/__init__.py to maintain functioning code. Previously these methods were in the `__init__.py` files of the various Azure Functions.
```
# Within __init__.py in NLPQueue or ProcessQueue
from shared_code import omop_helpers
concept_ids = omop_helpers.get_concept_from_concept_code(concept_code=139490008, vocabulary_id="SNOMED")
```
2. Updates `ProcessQueue/__init__.py` to include code to convert concept codes to conceptIDs on scan report upload. If a data dictionary defines a code vocabulary for a field, the new code now looks up each value's code and converts it to a valid and standard conceptID using `omop_helpers.get_concept_from_concept_code()`.
3. The newly looked up conceptID is temporarily saved to `ScanReportValue__conceptID`
4. After POSTing (and PKs for each value have been generated) the code resets the conceptID field back to -1 (default) by using HTTP PATCH via the API.

After consultation with the data team (Scott H), the following vocabs (defined at the top of ProcessQueue/__init__.py) are used when looking up concept codes -> conceptIDs:

```
vocabs = [
    "ABMS",
    "ATC",
    "HCPCS",
    "HES Specialty",
    "ICD10",
    "ICD10CM",
    "ICD10PCS",
    "ICD9CM",
    "ICD9Proc",
    "LOINC",
    "NDC",
    "NUCC",
    "OMOP Extension",
    "OSM",
    "PHDSC",
    "Read",
    "RxNorm",
    "RxNorm Extension",
    "SNOMED",
    "SPL",
    "UCUM",
    "UK Biobank",
]
```

Note that there is currently a small inefficiency in how the HTTP PATCH code works. I have raised task 39268 and assigned it to Joe Horder to improve this functionality.


NLPQueue/__init__.py
```diff
@@ -2,162 +2,21 @@
...
 import os
-import time
+
+from shared_code import omop_helpers
 
 api_url = os.environ.get('APP_URL')+"api/"
 api_header = {"Authorization": "Token {}".format(
     os.environ.get('AZ_FUNCTION_KEY'))}
 
-
```
simon-tarr, line 11: These methods moved to shared_code/omop_helpers.py

ProcessQueue/__init__.py
```diff
@@ -306,38 +345,131 @@ def main(msg: func.QueueMessage):
...
+                            row["code"]
+                            for row in data_dictionary
+                            if str(row["field_name"]) == str(name)
+                        ),
+                        None,
+                    )
+
+                    # If 'code' is in our vocab list, try and convert the ScanReportValue (concept code) to conceptID
+                    # If there's a faulty concept code for the vocab, fail gracefully and set concept_id to default (-1)
+                    if code in vocabs:
```
spco, line 377: Perhaps we should also add `and val_desc is None` to gracefully ignore anything where a valid code is supplied but the `value` has also been filled in - which indicates that the data dictionary is malformed or perhaps we have a code which matches the vocab list but shouldn't be mapped. Thoughts?

simon-tarr: Interesting idea.

I think maybe it would depend how the data dictionaries are used 'in the wild'. According to the current standards, there should only be 1 row in the data dictionary for a vocab-coded field; all values within that field belong to the vocabulary defined in the dictionary. If a data partner happened to supply what would technically be a value description but in reality be a field description, it would just get ignored.

On the other hand, a data provider could supply 1 row per value in the dictionary, even if it's coded under some kind of vocabulary. In this case I guess the question would be - would we want to bring across that value description into CCOM? In other words, would a value description be useful information to have (for human readability) even though the concept code (and therefore the conceptID is known?

PR #226, opened 2021-08-11
Comment out signup page from urls.py

vpnu:


spco:
@vpnu could you also remove lines 1026-9 in `views.py` to remove `SignUpView`, and remove the `SignUp` section in `README.md`? And while you're at it, remove this line rather than just commenting it out - we can go back through the history if we want to reinstate it later.

spco (approved):
Thanks!

PR #235, opened 2021-08-20
Dev/speed up rules

calmacx:


api/mapping/views.py
```diff
@@ -641,18 +642,18 @@ def get_context_data(self, **kwargs):
...
+        object_list = get_mapping_rules_list(self.get_queryset())        
         filtered_omop_table = self.kwargs.get("omop_table")
-        source_tables = list(set(
-            [
-                x.source_field.scan_report_table.name
-                for x in context['object_list']
-            ]
-        ))
-
+        source_tables =  list(set([
```
spco, line 647: Could you move the `filtered_omop_table` line down, as it is unrelated to the two lines around it, who _are_ related by `object_list`.

api/mapping/services_rules.py
```diff
@@ -393,7 +393,180 @@ def get_rules_by_scan_report_concept(scan_report_concept_id):
...
+                'destination_table':destination_table,
+                'destination_field':destination_field,
+                'source_table':source_table,
+                'source_field':source_field,
+                'term_mapping':term_mapping
+            })
+            
+    return rules
+
+def get_mapping_rules_json_batch(structural_mapping_rules):
```
spco, line 505: Please comment with how this is different to `_list`.

entrypoint.sh
```diff
@@ -9,4 +9,4 @@ mkdir staticfiles
 python /api/manage.py collectstatic
 
 # Set tmp dir to be in-memory for speed. Pass logs to stdout/err as Docker will expect them there
-gunicorn --worker-tmp-dir /dev/shm --timeout 600 --log-file=- --bind :8000 --workers 3 api.wsgi:application
+gunicorn --worker-tmp-dir /dev/shm --timeout 600 --log-file=- --bind :8000 --workers 3 api.wsgi:application --reload
```
calmacx, line 12: oh yeh @spco I had to app this to get the app to detect changes I have made to files in api.py, not sure why this became a problem for me, or if others had had this problem.

https://stackoverflow.com/questions/12773763/gunicorn-autoreload-on-source-change

spco: You're probably right, and there's hopefully no harm in it in the deployed setting either.

spco (commented):
Sorry, I added single comments through fat-thumbing. So as it stands, is this code used by the webapp, or does it still need plumbing in? From what I can see, `mappingrulesscanreport_list.html` would need editing.

api/mapping/views.py
calmacx: done

api/mapping/services_rules.py
calmacx: yeh good point, names are stupid. In the last commit I added more detail to the docstring and then called this function `get_mapping_rules_json` and removed the old `get_mapping_rules_json`

Summary:
* `get_mapping_rules_list`  --> gets a list of mapping rules that can be displayed by the template html 
* `get_mapping_rules_json`  --> first gets `get_mapping_rules_list`   and then converts it into a dictionary/json in the format needed by the ETLTool

calmacx:
As it stands this code can be used by the webapp with no problems.

On localhost, connected to the dev db, viewing the mapping rules page for table 56  on this branch takes 2secs
On devel branch it takes 30 secs.

PR #240, opened 2021-08-24
38912/upload json mapping rules

vpnu:


ProcessQueue/__init__.py
```diff
@@ -476,3 +479,71 @@ def main(msg: func.QueueMessage):
             field_ids = []
             field_names = []
             data = []
+    if parent_sr_id is not "None":
+        # Grab Parent Scan Report
+        parent_scan_report = requests.get(
+                    url=api_url
+                    + "scanreporttablesfilter/?scan_report="
+                    + str(parent_sr_id),
```
calmacx, line 487: pedantic comment - you don't need to fix :P !

it's more in line with pep-8 style to do this as one line, and maybe better using f-strings. i.e.
```
url=f"{api_url}scanreporttablesfilter/?scan_report={parent_sr_id}"
```

ProcessQueue/__init__.py
```diff
@@ -476,3 +479,71 @@ def main(msg: func.QueueMessage):
...
+
+        print("PARENT SCAN REPORT TABLES:",response_parent)
+        print("CHILD SCAN REPORT TABLES",response_child)
+
+        # Find differences between JSON?
+        # difference=diff(response_parent, response_child)
+        # print("JSON DIFF",difference)
+
+        # Find differences of two lists?
+        for element in range(len(response_parent)):
```
calmacx, line 512: pedantic comment ...

can you just not iterate over the element instead of the index?
i.e.
```
for element in response_parent:
      parent_tables.append(element["name"])
```
or
parent_tables = [
element['name']
for element in response_parent
]

ProcessQueue/__init__.py
```diff
@@ -476,3 +479,71 @@ def main(msg: func.QueueMessage):
...
+            parent_tables.append(response_parent[element]["name"])
+            parent_table_ids.append(response_parent[element]["id"])
+        for element in range(len(response_child)):
+            child_tables.append(response_child[element]["name"])
+        # Check if tables are the same
+        if parent_tables==child_tables:
```
calmacx, line 518: is this safe? i dont know if the responses are ordered - by that i mean, if the elements in the two lists are definitely going to be in the same order.

`[1,2,3,4] == [1,2,4,3]` : `False` when you might want this to be `True` as the lists are different but the elements are the same?

`sorted(parent_tables) == sorted(child_tables)` would be the solution to that (?)

ProcessQueue/__init__.py
```diff
@@ -476,3 +479,71 @@ def main(msg: func.QueueMessage):
             field_ids = []
             field_names = []
             data = []
+    if parent_sr_id is not "None":
```
calmacx, line 482: meant to be None, rather than "None" ?

calmacx: ahh no, it comes from some trick you do later with `azure_dict['parent_SR_id'] = "None"`

ProcessQueue/__init__.py
spco: I think the wider question is whether this is the right thing to be doing. We have a special case where `sorted(parent_tables) == sorted(child_tables)` but I don't know that it's all that useful to pick it out. Ultimately, whether or not we have absolutely the same or not, we only want the intersection of tables and intersection of fields within tables (and values within fields). So I don't think we need to spend time and code on the very special case where they're identical in every respect - that will be rare. Instead, just trim down to the intersection of the 2 scan reports (so where the fields in tables are the same, and where the values inside those fields are the same) and process those. Whether that's the whole table or not doesn't really matter.

spco: in sum (sorry for the abuse of all sorts of notations) :
```
for table in intersection(parent_tables, child_tables):
    for field in intersection(parent_tables[table].fields, child_tables[table].fields):
        copy field concepts
        for value in intersection(parent_tables[table].fields[field].values, child_tables[table].fields[field].values):
            copy value concepts
```

ProcessQueue/__init__.py
vpnu: @calmacx yeah I was looking for a better solution for that but I iterate over the index because response_parent is a JSON dict and you can't pass a dictionary as a list indice

ProcessQueue/__init__.py
spco: For my education, is there any difference between the `scanreporttablesfilter/?scan_report=10` and `scanreporttables/10`? Either in functionality or performance?

calmacx: @shahzadmumtaz22  correct me if i'm wrong, but the `scanreporttablesfilter` could be better to use (performance wise) as you're able to also specify which `fields` you need returning. The `scanreporttables` api will return everything (the full serialised model) for that report.

E.g. you can do (?) something like:
```
scanreporttablesfilter/?scan_report=10?fields=id,name
```
which just returns a json with the `id` and `name`, rather than all the junk about date_created and other fields you don't need etc. 

No idea if there's any difference if they're used as in @spco's comment.

spco: Ok, so the choice of which is "right" will depend on what we use from the response. Ta.

shahzadmumtaz22: @vpnu 
image

http://localhost:8080/api/scanreporttablesfilter/?scan_report=1 (this returns details of all the tables for a scanreport that has an id=1 (this is scan report id, a field that is in the scan_report_table entries to link back to scan_report). 

http://http://localhost:8080/api/scanreporttables/10 (this returns a table that has an id value=10 in the scanreporttable contents). In an api when you use **/api/tablename/primarykeyvalue**, a primarykeyvalue is usually an id for a given table whereas when you use **filter/?filter_field=[]** whereas filter_field is usually a field other than the primary key field. 

@calmacx You are right we can specify returning fields (as we did earlier for couple of endpoints) to make it cleaner when using it for any purpose. 

@calmacx , @spco and @vpnu  I have update all serializers to allow users to specify returning fields for all API endpoints (except for OMOP tables and if required the same can be replicated to those as well). 
image

ProcessQueue/__init__.py
calmacx: ah yeh, you probably want
```
for key,element in response_parent.items()
```
or 
```
for element in response_parent.values()
```
when iterating over a json/dict

PR #241, opened 2021-08-24
Merge test to prod 240821

spco:


ProcessQueue/readme.md
```diff
@@ -1,9 +1,13 @@
...
 
 ## Triggering the function
 When a White Rabbit Scan Report file is uploaded on the site, it is saved into Azure Blob Storage and a message is send to the storage Queue, which triggers the ProcessQueue function
 
 The function downloads the file from blob storage into a BytesIO object and iterates through the file using openpyxl and saves the tables, fields and values to the relevant django models.
 
+## Note to developers
+To run this function locally, copy `local.settings.json` file from 
+`Teams Software Team -> files -> env files -> local.settings.json` to the root of project. Ensure that it contains 
+all the variables from `sample-local.settings.json` from the repository.
```
calmacx, line 13: actually confused why we have this here and then what seems to be some type of copy of this here: https://co-connect.github.io/docs/AzureFunctions/processQueue/

api/api/settings.py
```diff
@@ -141,6 +142,8 @@
 
 USE_TZ = True
 
+CORS_ORIGIN_ALLOW_ALL = True
+
```
calmacx, line 146: is this safe for production? 

my amateur knowledge of enabling CORS is you sometimes need to do it when making API calls on localhost as you get a CORS error, which you wouldn't normally get when deployed on a domain.

calmacx (approved):

spco:
Not sure why the `build` is failing, I think because it still thinks it's merging into `devel` rather than `master`. Not certain how I can force it to retry correctly as yet.

api/api/settings.py
calmacx: https://stackoverflow.com/questions/39042799/cors-localhost-as-allowed-origin-in-production/39083377

shahzadmumtaz22: @calmacx, you are right that CORS are set up to allow API calls from other domains.

image

@spco Right now the CORS are set up to allow all domains, I have not tried it on production systems yet but I think they can be set up for a specific domain as shown in Figure.

Details of setting up CORS for Django App is available here: http://www.srikanthtechnologies.com/blog/python/enable_cors_for_django.aspx

spco: I think this needs further testing before we deploy any fix to prod, and I am minded to deploy this to prod as it stands. Am I right that we are still protected by the fact that all API calls require the secret token anyway, and we're not depending on cookies? But this would add further protection. Can you confirm my understanding is right?

shahzadmumtaz22: @spco Yes your understanding is correct that all API calls are protected by token authentication anyway.

PR #246, opened 2021-08-27
Robust uploads

spco:
This PR improves the ProcessQueue function in a number of ways to make it more transparent, robust, and helpful. Including:

- Checking the message dequeue count - if this isn't the first time then we should abort (to avoid reprocessing a broken scan report a second time, and generating up to 5 of each table needlessly). If it's the second time of seeing the message, we edit the dataset name to append `FAILED: ` to visibly mark the dataset on the scan report list page.
- Replacing of a number of variables with generic names like `response` that get reused multiple times with more descriptive set of individual variables like `fields_response`, `values_response`.
- Raising `HTTPError`s when POSTs fail, in order to surface some useful debug information.
- reducing the indenting of the second half of the function by reording if-else.
- Refactoring the helper function `process_scan_report_sheet_table()` to return all values in column a then all in column b (rather than all in row 1, then row 2) because not all columns have the same number of rows, and it's bonkers to reason about it in a row-wise manner :)
- Pythonising a number of loops to avoid  unnecessary`enumerate()` calls.

Possibly still to do - add more debug information to the `HTTPError` raisings, to show the failed message contents.


ProcessQueue/__init__.py
```diff
@@ -97,7 +106,7 @@ def process_scan_report_sheet_table(sheet):
 
 def main(msg: func.QueueMessage):
     logging.info("Python queue trigger function processed a queue item.")
-
+    print(datetime.utcnow().strftime("%H:%M:%S.%fZ"))
```
calmacx, line 109: print check left by mistake

ProcessQueue/__init__.py
```diff
@@ -128,22 +131,50 @@ def main(msg: func.QueueMessage):
             "pop_receipt": msg.pop_receipt,
             "dequeue_count": msg.dequeue_count,
         }
-    )
-
+    
+    print("message:", type(message), message)
```
calmacx, line 135: likewise

ProcessQueue/__init__.py
```diff
@@ -128,22 +131,50 @@ def main(msg: func.QueueMessage):
...
+        dequeue_response = requests.patch(
+            url=f"{api_url}scanreports/{body['scan_report_id']}/",
+            data=json_data, 
+            headers=headers
+        )
+    if msg.dequeue_count > 1:
+        raise Exception('dequeue_count > 1')
+
+    # Otherwise, this must be the first time we've seen this message. Proceed.
+    print("Get blobs", datetime.utcnow().strftime("%H:%M:%S.%fZ"))
```
calmacx, line 168: maybe you are wanting all these print statements for debugging anyway(?)

ProcessQueue/__init__.py
```diff
@@ -159,320 +190,343 @@ def main(msg: func.QueueMessage):
...
-        # If the value in the first column (i.e. the Table Col) is not blank;
-        # Save the table name as a new entry in the model ScanReportTable
-        # Checks for blank b/c White Rabbit seperates tables with blank row
-        for cell in row_cell:
-            if cell.value:
-                name = ws.cell(row=i, column=1).value
-                if name not in table_names:
-                    table_names.append(name)
+    # Skip header row
+    for cell in fo_ws['A'][1:]:
```
calmacx, line 202: hmm not sure why this `'A'` is hardcoded?  Just making sure it wasn't from testing.

ProcessQueue/__init__.py
```diff
@@ -159,320 +190,343 @@ def main(msg: func.QueueMessage):
...
 
-        if idx >= len(table_ids):
+        if table_idx >= len(table_ids):
             continue
         # If fraction empty or fraction unique is empty set to 0(decimal)
-        if not (ws.cell(row=i, column=8).value):
-            ws.cell(row=i, column=8).value = 0.0
-        if not (ws.cell(row=i, column=10).value):
-            ws.cell(row=i, column=10).value = 0.0
+        if not (fo_ws.cell(row=i, column=8).value):
```
calmacx, line 294: Dont really get all these lines. Earlier there was a change to iterate over the cells, and then `row_cell.value` was used. I guess this is because a different column is being grabbed? 

Not sure if my comments makes sense or not. Just double checking there's not a mistake or a more readable way of doing this. Im sure it's fine as the previous code worked.

ProcessQueue/__init__.py
```diff
@@ -159,320 +190,343 @@ def main(msg: func.QueueMessage):
...
-            "description_column": str(ws.cell(row=i, column=3).value),
-            "type_column": str(ws.cell(row=i, column=4).value),
-            "max_length": ws.cell(row=i, column=5).value,
-            "nrows": ws.cell(row=i, column=6).value,
-            "nrows_checked": ws.cell(row=i, column=7).value,
-            "fraction_empty": round(ws.cell(row=i, column=8).value, 2),
-            "nunique_values": ws.cell(row=i, column=9).value,
-            "fraction_unique": round(ws.cell(row=i, column=10).value, 2),
-            "flag_column": str(ws.cell(row=i, column=11).value),
+            "name": str(fo_ws.cell(row=i, column=2).value)[:63],
```
calmacx, line 304: We don't throw an error or anything if the name is too long, just truncate and will be fine? I guess this is just the name so should be (?)

ProcessQueue/__init__.py
```diff
@@ -159,320 +190,343 @@ def main(msg: func.QueueMessage):
...
+
+        concept_id_data = [{
+                "nlp_entity": None,
+                "nlp_entity_type": None,
+                "nlp_confidence": None,
+                "nlp_vocabulary": None,
+                "nlp_processed_string": None,
+                "concept": concept["conceptID"],
+                "object_id": concept["id"],
+                "content_type": 17,
```
calmacx, line 492: `17` seams really random, i guess it is. Is it safe to assume it's the same ID across all databases (I think so), or would be the same if a completely new instance of this webtool was started?

https://stackoverflow.com/questions/12716970/get-contenttype-id-in-django-for-generic-relation
like:
```
"content_type": ContentType.objects.get(model='scanreportvalue').id
```

calmacx (approved):
few comments, but dont think any are really super important. Just checks that it is what was intended

ProcessQueue/__init__.py
spco: I'm leaving these in for now in the (?vain) hope they will give us some useful guidance on timings in production.

ProcessQueue/__init__.py
spco: `'A'` is selecting the first column to grab all the table names.

ProcessQueue/__init__.py
spco: Good catch - this wasn't meant to go into production! Thanks.

ProcessQueue/__init__.py
spco: I've written a helper function `default_zero()` to neaten this up - I feel like there's possibly an easier way, but I can't see it right now.

spco:
@calmacx are you happy I've addressed everything before I merge?

ProcessQueue/__init__.py
calmacx: ahhhh of course

calmacx (approved):

PR #253, opened 2021-09-02
Download SVG and PNG Image

calmacx:
## The problem...
* I wanted to convert an svg image to a PNG so people could download and view

 
## Attempts
* server side convert the svg into a png and download, confusing stackoverflows on this didn't work
* post the svg html string (using jquery to get it from its div element displayed on the page) to python backend, then in python convert to png and then return png data via ajax to download... didnt work as data is not encoded correctly in base64 
* call get/post to retrieve the image again from make_dag() function, but in png form (instead format='png', instead of format='svg')... same issues with base64 encoding. Maybe writing to a temp file that could be downloaded is better
	 

## Thoughts 

* this is actually a complex problem from googling similar questions, mostly because I'm having to use AJAX to do the posting (due to spinners) 
* React probably has solutions to this, and maybe I shouldnt waste more time on this when the page will be react-ified soon

### DIFF
--- a/api/mapping/services_rules.py
+++ b/api/mapping/services_rules.py
@@ -583,9 +583,9 @@ def download_mapping_rules(request,qs):
     return response
 
 
-def make_dag(data):
-    dot = Digraph(strict=True,format='svg')
-    dot.attr(rankdir='RL', size='8,5')
+def make_dag(data,format='svg'):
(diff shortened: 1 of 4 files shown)


spco:
I believe we decided to halt this - users can convert svg to png locally if they need. Happy if I close this?

calmacx:
yep!

PR #256, opened 2021-09-02
Extend ScanReportField.name and .description_column to 512 chars.

spco:
This requires a migration to be applied to DBs as it is deployed.

### DIFF
--- a/api/mapping/models.py
+++ b/api/mapping/models.py
@@ -326,11 +326,11 @@ class ScanReportField(BaseModel):
     )
 
     name = models.CharField(
-        max_length=64,
+        max_length=512,
     )
 
     description_column = models.CharField(
-        max_length=256,
+        max_length=512,
     )
 
     type_column = models.CharField(


calmacx (approved):

calmacx:
@spco I just remembered the other migration that we could do, which was in StructuralMapping rules to rename `concept` to `scan_report_concept`
https://github.com/CO-CONNECT/mapping-pipeline/blob/devel/api/mapping/models.py#L478-L479

Not sure if we want to do this now at get all migrations out of the way, I can do a separate PR for it (?), but need to be careful as I know it is used in several places for creating the json rules.

spco:
@calmacx are you aware of any other occurrences? I believe I've grepped them all.

calmacx:
@spco  think there's a few here:

https://github.com/CO-CONNECT/mapping-pipeline/blob/devel/api/mapping/services_rules.py#L413
this will be one as it's the `_id` of the object `concept` which has been changed to `scan_report_concept` .... this is why this was so confusing.

https://github.com/CO-CONNECT/mapping-pipeline/blob/devel/api/mapping/services_rules.py#L482
https://github.com/CO-CONNECT/mapping-pipeline/blob/devel/api/mapping/services_rules.py#L483

```
  scan_report_concept_id = rule.concept_id
  scan_report_concept = scan_report_concepts[rule.concept_id]
  concept_id = scan_report_concept.concept_id
```
--->
```
  scan_report_concept_id = rule.scan_report_concept_id
  scan_report_concept = scan_report_concepts[scan_report_concept_id]
  concept_id = scan_report_concept.concept_id
```
I _think_ that's right 

let me know if you want me to double check/test

spco:
Thanks @calmacx - I've added those changes, but if you can check the change it would be good.

spco:
I've pulled out the lengthening of the `ScanReportField.name` and `.description_column` into #257 just because I wanted to progress that and test out the TWINS dataset.

PR #261, opened 2021-09-08
Extend the maximum name and description_column lengths in the API to 512

spco:


spco:
This is an interesting bug - I hadn't realised that the serializers enact their own max lengths on some member fields. @shahzadmumtaz22 is there some way to make these serializers just take the value from the model definitions? I don't want to manually have to check that they're compatible - if it comes to it we can query the model for the lengths, but that seems backwards.

shahzadmumtaz22:
Hi @spco the field length in serializers for API are usually taken as automatic based on the length in the model except for a couple of cases where we need to allow_blank=True while making a post request and we explicity specify length of the field.

PR #269, opened 2021-09-14
React edit field

Samad198:
Added edit field page to react app

### DIFF
--- a/react-client-app/src/App.jsx
+++ b/react-client-app/src/App.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react'
 import { ChakraProvider } from "@chakra-ui/react"
 import styles from './styles'
 import DataTbl from './components/DataTbl'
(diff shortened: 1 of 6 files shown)


react-client-app/src/App.jsx
```diff
@@ -7,8 +7,10 @@ import MappingTbl from './components/MappingTbl';
 import FieldsTbl from './components/FieldsTbl';
 import TablesTbl from './components/TablesTbl';
 import EditTable from './components/EditTable';
+import EditField from './components/EditField';
 
 const App = ({ page }) => {
+    const [title, setTitle] = useState(page);
```
spco, line 13: What does this do, and why is it needed here but not for the 'Edit Table'?

Samad198: It is a function to set the page title. It was needed in the edit table page because the name of the field was being used in the title to make it something like "Update Field - identifier" rather than just "Update Field", so I needed to set the title dynamically inside the Edit table component.

PR #270, opened 2021-09-15
Shahzad api updates retrieve user details

shahzadmumtaz22:
Samuel Adejumo is working on the scanreport page and this update will help him to extract user details for each scan report using API endppoints (as in each scan report we only store id of the user and user name and other information are usually stored in djanog user related table).

### DIFF
--- a/api/mapping/serializers.py
+++ b/api/mapping/serializers.py
@@ -73,7 +73,7 @@ class DrugStrengthSerializer(serializers.ModelSerializer):
 class UserSerializer(serializers.ModelSerializer):
     class Meta:
         model = User
-        fields = '__all__'
+        fields = ('id','username')
         
 class ScanReportSerializer(DynamicFieldsMixin,serializers.ModelSerializer):
     class Meta:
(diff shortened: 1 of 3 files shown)


spco:
Is it worth restricting the fields that can be queried? It would be preferable if you can't pull out the email addresses for example - that is something only users with admin privileges should be able to see, but using this API you can view that if you have any API key (which an unprivileged developer might have, for example, who doesn't also have admin privileges in django)

shahzadmumtaz22:
> Is it worth restricting the fields that can be queried? It would be preferable if you can't pull out the email addresses for example - that is something only users with admin privileges should be able to see, but using this API you can view that if you have any API key (which an unprivileged developer might have, for example, who doesn't also have admin privileges in django)

Hi @spco 

Good pick up. 

I just have discussions with Samuel Adejumo and he said he will only need username and id. I have made changes and now it will only return username and id through an API endpoint.

spco:
Perfect, thanks! Are you happy for me to merge this now?

PR #277, opened 2021-09-24
React code improvements

Samad198:
Changed methods of adding and deleting concepts in field and values tables to use functions shared by both components


spco (commented):
Looks like some really useful refactoring here. Are you happy for this to be merged?

PR #279, opened 2021-09-24
Revert v1.1 merge as it's not starting the container correctly.

spco:


spco:
Note that the migration of name and description_column has still been applied, and the relevant env vars are still updated.

PR #284, opened 2021-10-07
React scanreport status

Samad198:
Added status dropdown to scanreports table

### DIFF
--- a/api/api/react.py
+++ b/api/api/react.py
@@ -5,5 +5,5 @@ def react(request):
     return {
         "a": os.environ.get("COCONNECT_DB_AUTH_TOKEN"),
         "u": os.environ.get("CCOM_APP_URL"),   
-        "status":[{'id': year[0], 'label': year[1]} for year in Status.choices]
+        "status":[{'id': id, 'label': label} for id, label in Status.choices]
     }


react-client-app/src/components/ScanReportTbl.jsx
```diff
@@ -19,6 +19,8 @@ const ScanReportTbl = (props) => {
     const [authorFilter, setAuthorFilter] = useState("All");
     const [title, setTitle] = useState("Scan Reports Active");
     const [expanded, setExpanded] = useState(false);
+    const statuses = ["UPINPRO", "UPCOMPL", "UPFAILE", "PENDING", "INPRO25", "INPRO50", "INPRO75", "COMPLET", "BLOCKED"]
```
spco, line 22: Is there any way to grab these (and the contents of `mapStatus()` below) from the django model rather than hardcoding them here? Better for us to only have to change it in one place.

api/api/react.py
```diff
@@ -1,7 +1,9 @@
 from django.conf import settings
 import os
+from mapping.models import Status
 def react(request):
     return {
         "a": os.environ.get("COCONNECT_DB_AUTH_TOKEN"),
-        "u": os.environ.get("CCOM_APP_URL")     
+        "u": os.environ.get("CCOM_APP_URL")   
+        "status":[{'id': year[0], 'label': year[1]} for year in Status.choices]
```
spco, line 8: Please remove the `year` here which is not relevant. Either 
`[{'id': choice[0], 'label': choice[1]} for choice in Status.choices]` or (even better) `[{'id': id, 'label': label} for id, label in Status.choices]`

PR #322, opened 2021-11-08
Save value ids and names in dictionary

vpnu:
**What we have**
_Input_: Field names of currently uploaded scan report

**What we need**
Concept IDs for field names that have already been mapped

**Could be a helpful endpoint:**
Add a filter for content_type: 17 (values) or 15(fields)
Get all the fields with existing mappings
Get all values with existing mappings
_example_:
  Get concept for ScanReportField with ids in field_list[]
  api/scanreportconceptsfilter/object_id__in={field_list}&&content_type=15

1st Method:
1. Find the fields with the same name in the ScanReportField model
2. If fields have the same name use the id and search if a mapping exists in ScanReportConcept table(object_id=field_id)
3. If an entry is found then get the conceptID
4. Create a new ScanReportConcept entry for the new fields:
    new_concept={  ...,
                    "concept": conceptID(step 3),
                    "object_id": field_id,
                    "content_type": 15,
                }

2nd Method:

1. For all records in ScanReportConcept find the fields and their conceptIDs-> dict= field_id:concept
2. For each id (dict key) query ScanReportField table for the field names 
3. We now have field ids: names: concepts
4. Create a new ScanReportConcept entry for the new field ids

### DIFF
--- a/ProcessQueue/__init__.py
+++ b/ProcessQueue/__init__.py
@@ -172,23 +172,52 @@ def get_existing_concepts(name_ids,content_type,api_url,headers):
     ids=", ".join(ids)
     names=list(name_ids.keys())
     names=", ".join(names)
+
     get_field_names=requests.get(
         url=f"{api_url}scanreportfieldsfilter/?id__in={ids}&name__in={names}",
         headers=headers,
     )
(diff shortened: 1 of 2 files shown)


vpnu:
> @vpnu can this be closed now?

Yep all of this code is old

PR #329, opened 2021-11-16
started phenotyping queries

Samad198:
Output with 8507 hardcoded as the concept id is 
image

### DIFF
--- a/api/mapping/management/commands/create_concept.py
+++ b/api/mapping/management/commands/create_concept.py
@@ -6,32 +6,43 @@ from mapping.services_rules import (
     remove_mapping_rules,
     find_existing_scan_report_concepts
 )
-from django.http import HttpRequest
+
 
 
 class Command(BaseCommand):
     help = 'create rules from existing concepts found for a given table id'
 
     def handle(self, *args, **options):
-
+        use_archived = False
         concept_id = 8507 # id for male
         scan_report_concepts = ScanReportConcept.objects.filter(concept=concept_id) 
-        
         finalList = []
-        #print(scan_report_concepts.values())
+        
         for scan_report_concept in scan_report_concepts:
             if (scan_report_concept.content_type_id==17):
                 # If it is for a value
-                scan_report_values = ScanReportValue.objects.filter(pk=scan_report_concept.object_id)
-                scan_value = scan_report_values[0]
-                scan_report_fields = ScanReportField.objects.filter(pk=scan_value.scan_report_field_id)
-                scan_field = scan_report_fields[0]
-                finalList.append((scan_field.name ,scan_value.value,scan_value.value_description))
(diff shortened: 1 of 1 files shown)


api/mapping/management/commands/create_concept.py
```diff
@@ -0,0 +1,38 @@
...
+        #print(scan_report_concepts.values())
+        for scan_report_concept in scan_report_concepts:
+            if (scan_report_concept.content_type_id==17):
+                # If it is for a value
+                scan_report_values = ScanReportValue.objects.filter(pk=scan_report_concept.object_id)
+                scan_value = scan_report_values[0]
+                scan_report_fields = ScanReportField.objects.filter(pk=scan_value.scan_report_field_id)
+                scan_field = scan_report_fields[0]
+                finalList.append((scan_field.name ,scan_value.value,scan_value.value_description))
+            if (scan_report_concept.content_type_id==15):
```
spco, line 30: use `elif` for clarity

api/mapping/management/commands/create_concept.py
```diff
@@ -0,0 +1,38 @@
...
+)
+from django.http import HttpRequest
+
+
+class Command(BaseCommand):
+    help = 'create rules from existing concepts found for a given table id'
+
+    def handle(self, *args, **options):
+
+        concept_id = 8507 # id for male
```
spco, line 17: please set this as an input argument rather than hardcoded

spco (commented):
Great stuff in general, a few comments to tidy it up.

The remaining piece of logic I'd like (sorry, hadn't mentioned this before now!) is that we should have an optional argument to indicate whether to use all scan reports, or to ignore mappings from archived reports. Could you add this please? Call it `use_archived`, defaulting to `False`, and if `False` then perform a check or filter to only extract mappings associated to fields or values in ScanReports where the column `hidden=False`.

If you could also add a few comments to really make the logic crystal clear that would be useful.

api/mapping/management/commands/create_concept.py
Samad198: Done

api/mapping/management/commands/create_concept.py
Samad198: Done

Samad198:
> Great stuff in general, a few comments to tidy it up.
> 
> The remaining piece of logic I'd like (sorry, hadn't mentioned this before now!) is that we should have an optional argument to indicate whether to use all scan reports, or to ignore mappings from archived reports. Could you add this please? Call it `use_archived`, defaulting to `False`, and if `False` then perform a check or filter to only extract mappings associated to fields or values in ScanReports where the column `hidden=False`.
> 
> If you could also add a few comments to really make the logic crystal clear that would be useful.

Done

PR #330, opened 2021-11-17
make sure the id is unique for the rules json

calmacx:
Bug spotted by Erum in which mappings in the JSON were being overwritten because of the use of a dictionary structure


spco:
Looks fine to me - are you satisfied with your own testing?

calmacx:
yep, it works fine and stops the replacement issue

PR #334, opened 2021-11-22
removed current user check

Samad198:
removed current user check on scan reports to be able to archive
image


react-client-app/src/components/ScanReportTbl.jsx
```diff
@@ -336,10 +336,6 @@ const ScanReportTbl = (props) => {
                                     <Link href={"/scanreports/" + item.id + "/assertions/"}><Button variant="green">Assertions</Button></Link>
                                 </Td>
                                 <Td textAlign="center">
-                                    {currentUser &&
-                                        <>
-                                            {currentUser == item.author.username &&
-                                                <>
```
spco, line 342: Could you please modify the indenting of the inner statements as well.

Samad198: Done

PR #339, opened 2021-11-26
Async posts

spco:


spco:
We add in a new environment variable `CHUNK_SIZE` which should be set in the deployed environments to `6`

PR #354, opened 2022-01-25
Creation type model

Samad198:
Moved creation_type field from MappingRule to ScanreportConcept, and changed codebase to accomodate the change


spco (commented):
Please add a changelog entry into 1.4.0-beta including noting this requires a migration.

PR #356, opened 2022-01-26
Feature/projects and datasets

dcl10:
# Add Project and Dataset models

- Created `Dataset` model and created a Foreign Key relation to it from `ScanReport`
- Created `Project` model with many to many relation to Users and Datasets
- Created django management command to associate orphaned scan reports to a Dataset and via a Dataset, a project.

### DIFF
--- a/api/mapping/models.py
+++ b/api/mapping/models.py
@@ -2,6 +2,7 @@
 To come
 """
 
+import uuid
 from django.conf import settings
 from django.db import models
 from django.db.models.constraints import UniqueConstraint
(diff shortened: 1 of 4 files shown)


api/mapping/models.py
```diff
@@ -495,15 +360,49 @@ class NLPModel(models.Model):
...
+    members = models.ManyToManyField(
+        settings.AUTH_USER_MODEL, related_name="projects", related_query_name="project"
+    )
+
+    class Meta:
+        verbose_name = "Project"
+        verbose_name_plural = "Projects"
+
+    def __str__(self) -> str:
+        return self.name
```
dcl10, line 408: Review here @spco

api/mapping/models.py
```diff
@@ -190,47 +142,44 @@ class ScanReport(BaseModel):
...
+    parent_dataset = models.ForeignKey(
+        "Dataset",
+        on_delete=models.CASCADE,
+        related_name="scan_reports",
+        related_query_name="scan_report",
+        null=True,
+        blank=True,
     )
-    
+
```
dcl10, line 182: Review here @spco. I put the note to rename it in future to a less confusing name as well as changing the one that is already called `dataset` since that one is really just the name and `name` is for something else.

PR #362, opened 2022-01-28
Upload SR changes

dcl10:
## Please merge #370 before this PR
# Changes
- Alter `ScanReport` form to have a field for selecting a `parent_dataset`
- Change the form labels 
  - from "Dataset name", corresponding to the `dataset` field in `ScanReport`, to "Scan Report name"
  - "Dataset" denotes the `parent_dataset` field in `ScanReport`
- `ScanReportFormView` now adds the `parent_dataset` field to the `ScanReport` instance it creates

Screenshot 2022-01-28 at 15 16 44

### DIFF
--- a/api/mapping/forms.py
+++ b/api/mapping/forms.py
@@ -45,7 +45,7 @@ class ScanReportForm(forms.Form):
 
     class Meta:
         model = ScanReport
-        fields = ("data_partner", "dataset", "scan_report_file", "parent_dataser")
+        fields = ("data_partner", "dataset", "scan_report_file", "parent_dataset")
 
     def clean_data_dictionary_file(self):
 
--- a/api/mapping/forms.py
+++ b/api/mapping/forms.py
@@ -5,7 +5,7 @@ from django.contrib.auth.models import User
 from django.core.exceptions import ValidationError
 from django.forms.models import ModelChoiceField
 
-from mapping.models import DataPartner, ScanReportField, ScanReport
+from mapping.models import DataPartner, Dataset, ScanReportField, ScanReport
 import openpyxl
 import csv
 from io import BytesIO, StringIO
@@ -24,7 +24,8 @@ class ScanReportForm(forms.Form):
         widget=forms.Select(attrs={"class": "form-control"}),
     )
     dataset = forms.CharField(
-        label="Dataset name", widget=forms.TextInput(attrs={"class": "form-control"})
+        label="Scan Report name",
+        widget=forms.TextInput(attrs={"class": "form-control"}),
     )
     scan_report_file = forms.FileField(
         label="WhiteRabbit ScanReport",
@@ -36,10 +37,15 @@ class ScanReportForm(forms.Form):
         widget=forms.FileInput(attrs={"class": "form-control"}),
         required=False,
     )
+    parent_dataset = ShowNameChoiceField(
+        label="Dataset",
+        queryset=Dataset.objects.order_by("name"),
+        widget=forms.Select(attrs={"class": "form-control"}),
+    )
 
     class Meta:
         model = ScanReport
-        fields = ("data_partner", "dataset", "scan_report_file")
+        fields = ("data_partner", "dataset", "scan_report_file", "parent_dataser")
 
     def clean_data_dictionary_file(self):
 
--- a/api/mapping/models.py
+++ b/api/mapping/models.py
@@ -42,9 +42,9 @@ class BaseModel(models.Model):
     To come
     """
 
-    created_at = models.DateTimeField(auto_now_add=True,)
+    created_at = models.DateTimeField(auto_now_add=True)
 
-    updated_at = models.DateTimeField(auto_now=True,)
+    updated_at = models.DateTimeField(auto_now=True)
 
     class Meta:
         abstract = True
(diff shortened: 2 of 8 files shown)


dcl10:
@calmacx Before merging I might need to rebase this branch after @spco finishes PR #356

calmacx:
cheers @dcl10 , will start taking a look, and ignore merge conflicts etc. for now

api/mapping/forms.py
```diff
@@ -36,10 +37,15 @@ class ScanReportForm(forms.Form):
...
+    parent_dataset = ShowNameChoiceField(
+        label="Dataset",
+        queryset=Dataset.objects.order_by("name"),
+        widget=forms.Select(attrs={"class": "form-control"}),
+    )
 
     class Meta:
         model = ScanReport
-        fields = ("data_partner", "dataset", "scan_report_file")
+        fields = ("data_partner", "dataset", "scan_report_file", "parent_dataser")
```
calmacx, line 48: `parent_dataser` --> typo?

api/mapping/models.py
```diff
@@ -354,3 +366,43 @@ class NLPModel(models.Model):
...
     def __str__(self):
         return str(self.id)
+
+
+class Dataset(BaseModel):
+    """
+    Model for datasets which contain scan reports.
+    """
+
+    id = models.UUIDField(primary_key=True, editable=False, default=uuid.uuid4)
```
calmacx, line 376: why do we need an `id` field that is `uuid4` ? What's wrong the the `django` default (from `BaseModel`) `id` field?

api/mapping/models.py
```diff
@@ -354,3 +366,43 @@ class NLPModel(models.Model):
...
+    def __str__(self) -> str:
+        return self.name
+
+
+class Project(BaseModel):
+    """
+    Model for projects which are made up of datasets.
+    """
+
+    id = models.UUIDField(primary_key=True, editable=False, default=uuid.uuid4)
```
calmacx, line 394: same comment

api/mapping/forms.py
dcl10: haha yes

api/mapping/models.py
dcl10: The default auto-increment integer is less safe than UUIDs. If/when you get hacked, having id as 1, 2, 3 makes discoverable by iteration. Using random UUIDs stops people requesting `sensitive_item` 1-infinity

api/mapping/models.py
```diff
@@ -354,3 +366,43 @@ class NLPModel(models.Model):
...
+    name = models.CharField(max_length=100)
+    # `projects` field added by M2M field in `Project`
+    # `scan_reports` field added by FK field in `ScanReport`
+
+    class Meta:
+        verbose_name = "Dataset"
+        verbose_name_plural = "Datasets"
+
+    def __str__(self) -> str:
+        return self.name
```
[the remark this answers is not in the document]
dcl10: `self.name` is more readable than `self.id` especially when you use UUIDs.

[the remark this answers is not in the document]
spco: I'd be interested to know why this was decided (before my time). Calum, can you remember the argument (or who was in it)?

dcl10:
Hi @calmacx I'm very sorry. I've just noticed it's also showing changes that @spco already merged. That's because I branched this PR off another branch, now merged. But the thing is getting confused. I'll go and comment the actual changes for this PR in particular

api/mapping/models.py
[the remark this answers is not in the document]
calmacx: @vpnu might remember, it might have been less of an argument and more we were just told to change them all to `str(self.id)` without much explanation

api/mapping/models.py
calmacx: makes sense. Was/is there a plan to migrate existing models or just these, out of interest? Just had a think and dont think that'd be a problem at all for the API / rules generation for the mapping json file - often we used the existing 'id' just as a string key

dcl10: There's not a plan so far. Something to think about though.

api/mapping/models.py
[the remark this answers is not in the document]
vpnu: Yeah it was just decided, don't remember being told why either. Before that we used to have self.name instead

dcl10:
Hi @calmacx for some reason the black check is failing here but when I run it on the "bad" file locally everything is fine. I've tried pulling and rebasing to devel and neither had an effect. And the `models.py` here is identical to my local copy.
Either way I sorted out the ID fields

calmacx:
hi @dcl10, yeh im not too sure, first time i've seen this 'black job'.

E.g. this:
image

Is it failing because the original is the correct formatting for PEP8 (which i assume the job is checking against)?

dcl10:
> hi @dcl10, yeh im not too sure, first time i've seen this 'black job'.
> 
> E.g. this: image
> 
> Is it failing because the original is the correct formatting for PEP8 (which i assume the job is checking against)?

@calmacx That's a difference between black and pep8 I assume. But it won't prevent the code from working. But either way the problem I was having black on GH call me u for something different to the code on my computer. Turns out, it's because the job is checking out the wrong branch so it's not checking the right stuff atm. So you can ignore it for now.

spco:
Please rebase onto `devel` to have the fix from #371 included.

dcl10:
Can we review #370 first and merge it @calmacx , please? It will fix the still failing formatting here. Also that PR merges to this branch. This makes the important changes easier to see as well.

calmacx (commented):
ok looking good, will wait for approval until #370 is on devel and merged with that

PR #367, opened 2022-01-31
changed concept colours

Samad198:
Colours added
image

Through different colour blindness lenses

image

image

image


calmacx (approved):
uncontroversial :)

PR #370, opened 2022-02-01
Feature/project permissions

dcl10:
# Changes
- Added `ProjectListView` which shows all projects. No additional permissions required.
  - found at `/projects`
- Added `ProjectRetrieveView` which shows a specific project but only if the user is a member of that project.
  - found at `/projects/<int:pk>`
- Added permission class `CanViewProject` which checks if a user is a member of the project they are trying to view.

# Screenshots
### `/projects`
Screenshot 2022-02-02 at 09 33 06

### `/projects/<non-member project>`
Screenshot 2022-02-02 at 09 33 56

### `/projects/<member project>`
Screenshot 2022-02-02 at 09 36 19

### DIFF
--- a/api/mapping/views.py
+++ b/api/mapping/views.py
@@ -204,7 +204,7 @@ class ProjectListView(ListAPIView):
     serializer_class = ProjectSerializer
 
     def get_queryset(self):
-        return Project.objects.filter(members=self.request.user)
+        return Project.objects.all()
 
 
 class ProjectCreateView(CreateAPIView):
--- a/api/mapping/views.py
+++ b/api/mapping/views.py
@@ -201,9 +201,46 @@ class ProjectListView(ListAPIView):
     """
 
     permission_classes = []
+    serializer_class = ProjectSerializer
 
     def get_queryset(self):
         return Project.objects.filter(members=self.request.user)
+
+
+class ProjectCreateView(CreateAPIView):
+    """
+    API view to create a project.
+    """
+
+    permission_classes = []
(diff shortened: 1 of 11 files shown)


spco (commented):
In general looks good. Can you perhaps show this working in action before we merge to devel please? There are enough moving parts on this and all the following PRs for permissions that I'm sure we are going to miss things. Going through with demos is most likely to catch issues early. I'm rubbish at understanding whether you've done everything necessary in this PR.

api/mapping/views.py
```diff
@@ -988,11 +1015,7 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
         x = ScanReport.objects.get(pk=self.kwargs.get("pk"))
-        context.update(
-            {
-                "scan_report": x,
-            }
-        )
+        context.update({"scan_report": x})
```
calmacx, line 1018: odd that this doesnt fail the black job (?)

calmacx (approved):

calmacx:
LGTM... just will wait until tomorrow morning to demo as @spco  suggested before merge to devel branch ?

api/mapping/views.py
dcl10: The trailing comma at the end of line 993 was what caused it to spread out

spco:
All good by me now - will need to run the unorphan_scanreports command on `ccnet-dev-db` to correctly associate `Datasets` to `Projects`.

PR #374, opened 2022-02-02
Use formatted strings

spco:


api/mapping/views.py
```diff
@@ -914,18 +914,14 @@ def form_valid(self, form):
...
         scan_report = ScanReport.objects.create(
             data_partner=form.cleaned_data["data_partner"],
             dataset=form.cleaned_data["dataset"],
             parent_dataset=form.cleaned_data["parent_dataset"],
-            name=os.path.splitext(str(form.cleaned_data.get("scan_report_file")))[0]
-            + "_"
-            + dt
-            + rand
-            + ".xlsx",
+            name=f"{os.path.splitext(str(form.cleaned_data.get('scan_report_file')))[0]}_{dt}_{rand}.xlsx",
```
calmacx, line 924: i know this is legacy code, but the name getting is a bit messy, could be a function and better use `os.path.splitext`   when this is repeated for xlsx and csv files (scanreport would always be .xlsx? data dictionary would always be .csv?)

calmacx (approved):

api/mapping/views.py
spco: It could be moved into some function, although the `get()` will return different strings in the SR and DD case - the file names are not merely the same but with different extensions. Thoughts?

calmacx: ah yeh you want rand and dt to be the same

 so yeh just being pedantic thinking if that'll ever be a problem in the future if it's been hard coded that the scanreport has to be .xlsx (probably safe to assume), or the data dictionary is .csv (though it's just a name right?)

spco: Good point - please see latest commit.

calmacx (approved):

PR #375, opened 2022-02-02
Move datapartner to dataset

dcl10:
# Changes
- Add `data_partner` field to `Dataset`.
  - `NULL` is allowed for now but once the production data is migrated, this can be disallowed. Datasets must be owned by a Data Partner.
- Created `add_datasets_to_partner` management command.
  - Use to set datasets' data partners to that of the first scan report they contain, or to the one specified in the command line.

### DIFF
--- a/api/mapping/management/commands/add_datasets_to_partner.py
+++ b/api/mapping/management/commands/add_datasets_to_partner.py
@@ -30,7 +30,7 @@ class Command(BaseCommand):
         data_partner_name = options.get("data_partner")
         dataset_names = options.get("datasets")
 
-        print(f"Datasets will be added to Data Partner: {data_partner_name}")
+        print(f"Datasets will be added to Data Partner: {data_partner_name}.")
 
         if dataset_names:
             # Get only the specified datasets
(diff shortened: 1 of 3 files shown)


dcl10:
> > @dcl10 if this is a single PR, how do you intend to do the django migration? If I checkout this branch and make and do the associated migration, I'll lose the data in `ScanReport.data_partner` before I can run the `add_datasets_to_partner` command, no?
> 
> Good point. I'll put the `data_partner` back in `ScanReport`, but leave the other changes. Then do a new PR for getting rid of it.

@spco Done.

calmacx (approved):

calmacx:
Is a little unclear to me if you want this auto merged to devel right now or not (unsure about migrations), so i'll approve and let you guys do what is best

api/mapping/management/commands/add_datasets_to_partner.py
```diff
@@ -0,0 +1,54 @@
...
+            orphaned_datasets = Dataset.objects.filter(name__in=dataset_names)
+        else:
+            # Get only the datasets with no data partner
+            print("No datasets given. Fetching the unpartnered datasets...")
+            orphaned_datasets = Dataset.objects.filter(data_partner=None)
+
+        # Attach datasets to the data partner
+        print(f"Attaching datasets to Data Partner: {data_partner_name}")
+        for dataset in orphaned_datasets:
+            data_partner, _ = DataPartner.objects.get_or_create(name=data_partner_name)
```
spco, line 47: Please move this to line 53 as we only need to create this if we ever reach a `Dataset` without an `SR`. Otherwise, we don't want to create this default DP.

api/mapping/management/commands/add_datasets_to_partner.py
```diff
@@ -0,0 +1,54 @@
...
+            # Get only the specified datasets
+            print("Fetching the specified datasets...")
+            orphaned_datasets = Dataset.objects.filter(name__in=dataset_names)
+        else:
+            # Get only the datasets with no data partner
+            print("No datasets given. Fetching the unpartnered datasets...")
+            orphaned_datasets = Dataset.objects.filter(data_partner=None)
+
+        # Attach datasets to the data partner
+        print(f"Attaching datasets to Data Partner: {data_partner_name}")
```
spco, line 45: Also this doesn't really describe what's happening fully: if the datasets have an SR then they ignore `data_partner_name` and use their own.

PR #379, opened 2022-02-07
Visibility and viewers

dcl10:
# Important
Merge #382 into this branch before merging this PR.

# Changes
- Add 2 visibility choices for Scan Reports and Datasets
  - "Public" where a Scan Report or Dataset is available to users who are members of projects the Datasets and Scan Reports are in
  - "Restricted" where a Scan Report or Dataset can only be seen by those in the `viewers` fields
 - Added `visibility` field to `ScanReport` and `Dataset`
 - Added `viewers` many-to-many field to `ScanReport` and `Dataset`. Links these to the `User` model

### DIFF
--- a/api/mapping/models.py
+++ b/api/mapping/models.py
@@ -120,7 +120,10 @@ class ScanReportConcept(BaseModel):
     nlp_entity_type = models.CharField(max_length=64, null=True, blank=True)
 
     nlp_confidence = models.DecimalField(
-        max_digits=3, decimal_places=2, null=True, blank=True,
+        max_digits=3,
+        decimal_places=2,
+        null=True,
+        blank=True,
     )
 
     nlp_vocabulary = models.CharField(max_length=64, null=True, blank=True)
@@ -139,7 +142,9 @@ class ScanReportConcept(BaseModel):
 
     # save how the mapping rule was created
     creation_type = models.CharField(
-        max_length=1, choices=CreationType.choices, default=CreationType.Manual,
+        max_length=1,
+        choices=CreationType.choices,
+        default=CreationType.Manual,
     )
 
     def __str__(self):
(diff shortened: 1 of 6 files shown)


api/mapping/models.py
```diff
@@ -191,6 +196,19 @@ class ScanReport(BaseModel):
...
+        max_length=10,
+        choices=VisibilityChoices.choices,
+        default=VisibilityChoices.PUBLIC,
+    )
+
+    viewers = models.ManyToManyField(
+        settings.AUTH_USER_MODEL,
+        related_name="scanreport_viewings",
+        related_query_name="scanreport_viewing",
+        blank=True,
```
calmacx, line 209: ok so this will require some migrations / manipulation. on the current DB if we not going to allow already uploaded ScanReports to have these as fields (i.e. no `null=True` and `blank=True` ) ? 

If so, with is `viewers` allowed to be blank? That'd be no viewers of a scan report ?

dcl10: On this one there's no `null=True` because it doesn't work with ManyToManyField. This doesn't mean a default is needed though. But this is just the behaviour of this kind of field. Normally you would do something like `null=True` with `blank=True`.

RE: the migrations. They've already been done on model-test-db and they went just fine. We need to do the migrations on the dev db now.

calmacx: ahh interesting to know.

And for migrations, will you go ahead and do them yourself once this is merged? If so, please do.

dcl10: Sure I'll do them :)

dcl10: @calmacx Bit of a pickle. I can't do the migrations on dev. I don't seem to have permission. @spco should be able to help. Let's not merge this branch yet.

calmacx: thanks, im not too sure who can/cant. I also got a bit lost on our workflow here to be honest when it comes to migrations... if it'd be better to do migrations on dev first (and let people know) ; then deploy and migrate to test for the web-tool to be smoke tested

PR #377, opened 2022-02-04
React fixes

Samad198:
Scan report list and home pages are now fixed but are slower from querying datasets by id 1 by 1, Need to add a new endpoint to be able to batch query them

PR #383, opened 2022-02-09
added endpoint to batch query datasets

Samad198:


api/mapping/views.py
```diff
@@ -253,6 +253,19 @@ class DatasetListView(generics.ListAPIView):
...
+    """
+    API view to filter datasets by list of id's.
+    """
+
+    queryset = Dataset.objects.all()
+    serializer_class = DatasetSerializer
+    filter_backends = [DjangoFilterBackend]
+    filterset_fields = {
+        "id": ["in"],
+    }
```
dcl10, line 266: could you also add in field for filtering by project. So if we wanted, we could restrict the listed datasets to those on a specific project

react-client-app/src/components/ScanReportTbl.jsx
```diff
@@ -43,15 +43,15 @@ const ScanReportTbl = (props) => {
...
+        const authorIds = chunkIds(Object.keys(authorObject))
+        const datasetIds = chunkIds(Object.keys(datasetObject))
         const authorPromises = [];
         const datasetPromises = [];
         for (let i = 0; i < authorIds.length; i++) {
             authorPromises.push(useGet(`/usersfilter/?id__in=${authorIds[i].join()}`));
         }
         for (let i = 0; i < datasetIds.length; i++) {
-            datasetPromises.push(useGet(`/datasets/${datasetIds[i]}`));
+            datasetPromises.push(useGet(`/datasetsfilter/?id__in=${datasetIds[i]}`));
```
dcl10, line 54: Do you not need the `.join()` bit here too?

Samad198: Oh yeah. Good spot

api/mapping/views.py
calmacx: why are we adding this as a `generics.ListAPIView` rather than in the class serialisers?

i.e. not here: https://github.com/CO-CONNECT/mapping-pipeline/blob/devel/api/mapping/serializers.py

To be honest, this type of view looks better/simpler than how we've got other api endpoints setup and working (which involved creating a view as well as a serialiser)

Samad198: Yeah I think that's right

dcl10 (approved):

calmacx (approved):

spco:
Fixes the speed issues introduced in #377

PR #388, opened 2022-02-10
Unit tests for dataset and scan report perms

dcl10:
# Changes
- Added unit tests for Dataset permissions
- Added unit tests for Scan Report permissions


calmacx (approved):
Looks fine to me from scanning, can say i 100% understand but since it's a unit test there's no problem!

PR #392, opened 2022-02-15
Feature/dataset view permissions

dcl10:
# Changes
- Overridden `get_queryset` in `views.DatasetListView`
  - it will only return Datasets that are public or restricted Datasets that a suer is a viewer of
  - These will only be Datasets from Projects a user is a member of
- Added filter backend to `views.DatasetListView` from `views.DatasetFilterView`
- Removed `views.DatasetFilterView` and associated URL
- Replaced filter view URL calls with calls to `views.DatasetListView`
- Added unit tests for `views.DatasetListView`
- Added temporary fix so React code gets the right user token

### DIFF
--- a/api/mapping/test_views.py
+++ b/api/mapping/test_views.py
@@ -64,17 +64,19 @@ class TestDatasetListView(TestCase):
         )
         # Get the response data
         response_data = self.view(request).data
+        expected_objs = [
+            self.public_dataset1.id,
+            self.public_dataset2.id,
+            self.restricted_dataset.id,
+        ]
 
         # Assert user1 can only public_dataset1, public_dataset2
         # and restricted_dataset
+        self.assertEqual(len(response_data), len(expected_objs))
         for obj in response_data:
             self.assertIn(
                 obj.get("id"),
-                [
-                    self.public_dataset1.id,
-                    self.public_dataset2.id,
-                    self.restricted_dataset.id,
-                ],
+                expected_objs,
             )
 
         # Assert user1 can't see public_dataset3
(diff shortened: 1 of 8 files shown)


api/mapping/test_views.py
```diff
@@ -0,0 +1,161 @@
...
+            user=self.user1,
+            token=self.user1.auth_token,
+        )
+        # Get the response data
+        response_data = self.view(request).data
+
+        # Assert user1 can only public_dataset1, public_dataset2
+        # and restricted_dataset
+        for obj in response_data:
+            self.assertIn(
```
spco, line 71: Can we check some other way - if there is a mistake and user1 _cannot_ access `restricted_dataset` then this test would still (incorrectly) pass, right?

api/mapping/test_views.py
```diff
@@ -0,0 +1,161 @@
...
+        # Authenticate az_user
+        force_authenticate(
+            request,
+            user=az_user,
+            token=az_user.auth_token,
+        )
+        # Get the response
+        response_data = self.view(request).data
+        # Assert az_user can see all datasets
+        for obj in response_data:
```
spco, line 160: Does this positively show `response_data` contains all the datasets, or simply that `response_data` is a subset of all datasets?

api/mapping/test_views.py
```diff
@@ -64,17 +64,19 @@ def test_dataset_returns(self):
...
         response_data = self.view(request).data
+        expected_objs = [
+            self.public_dataset1.id,
+            self.public_dataset2.id,
+            self.restricted_dataset.id,
+        ]
 
         # Assert user1 can only public_dataset1, public_dataset2
         # and restricted_dataset
+        self.assertEqual(len(response_data), len(expected_objs))
```
spco, line 75: Rather than comparing length and asserting contents, could we rather more directly do 
```
assertEqual(sorted(expected_objs), sorted(response_data))
```
?

PR #393, opened 2022-02-17
Analyse concepts

vpnu:
Screenshot 2022-02-17 at 15 09 45

**New Modal for displaying Ancestors and Descendants.** 

- Name of the Button for the modal is "Analyse Rules" - do we want to change it to: _"See/Find Misaligned Mappings"_, _"Analyse Concepts"_, _"See Concept Hierarchy"_ or anything else?
Screenshot 2022-02-17 at 15 19 47

- Ancestors show in blue, Descendants in green (Used the co-connect colours for now)
- Only use rules with omop field containing "_concept_id" same as in Summary View, to keep rules from showing more than once.
- For concepts mapped to a field, link to the fields page using the table_id, with the link name being the corresponding field 
- For concepts mapped to a value, link to the values page using the field_id, with the link name being the field that value is in.
- Source Column will display the links in a single line with a space between (do we want to separate them more?)
- The size for the columns adjusts according to the text length

**Updated Modal to show min/max separation**
Screenshot 2022-03-07 at 11 40 57


api/mapping/services_rules.py
```diff
@@ -1,17 +1,19 @@
+from calendar import c
```
dcl10, line 1: I can't see where this is used. If it's not used, let's remove it
```suggestion
```

api/mapping/services_rules.py
```diff
@@ -841,3 +843,101 @@ def remove_mapping_rules(request, scan_report_id):
...
+            descendants = ConceptAncestor.objects.filter(ancestor_concept_id=rule)
+            ancestors = ConceptAncestor.objects.filter(descendant_concept_id=rule)
+
+            for descendant in descendants:
+                desc = descendant.descendant_concept_id
+                if (desc in all_mapping_rules) & (desc != rule):
+                    concept = Concept.objects.get(concept_id=desc)
+                    rule_name = Concept.objects.get(concept_id=rule).concept_name
+                    source_ids = (
+                        MappingRule.objects.filter(Q(concept__concept=desc))
```
dcl10, line 874: I think you don't need a `Q` object here
```suggestion
                        MappingRule.objects.filter(concept__concept=desc)
```

api/mapping/services_rules.py
```diff
@@ -841,3 +843,101 @@ def remove_mapping_rules(request, scan_report_id):
...
+    all_mapping_rules = MappingRule.objects.exclude(
+        scan_report_id=scan_report_id
+    ).values_list("concept__concept", flat=True)
+    data = []
+    descendant_list = []
+    ancestors_list = []
+
+    for rule in mapping_rules:
+
+        try:
```
dcl10, line 864: What kind of exceptions/errors happen when the `try` block fails?

api/mapping/services_rules.py
```diff
@@ -841,3 +843,101 @@ def remove_mapping_rules(request, scan_report_id):
...
+                        {
+                            "descendants": descendant_list,
+                            "ancestors": ancestors_list,
+                        }
+                    ],
+                }
+            )
+            descendant_list = []
+            ancestors_list = []
+        except:
```
dcl10, line 940: If this really does need to be in a `try`/`except` loop, could we put something in here to explain what went wrong please?

api/mapping/services_rules.py
```diff
@@ -841,3 +843,101 @@ def remove_mapping_rules(request, scan_report_id):
     rules = MappingRule.objects.all().filter(scan_report__id=scan_report_id)
 
     rules.delete()
+
+
+def analyse_concepts(scan_report_id):
```
dcl10, line 848: It seems like this code is doing a lot and some exception or error might occur (hence the `try`/`except` block). If the unit test set-up is in this branch, could we have a test/tests to show it gets the expected data and that it throws the expected exception/error when error condition is met. If that makes sense?

api/mapping/services_rules.py
```diff
@@ -845,6 +844,28 @@ def remove_mapping_rules(request, scan_report_id):
     rules.delete()
 
 
+def get_concept_hierarchy(rule, h_concept_id):
+    print(h_concept_id)
```
dcl10, line 848: Don't forget to remove this print statement.

dcl10 (dismissed):

api/mapping/forms.py
```diff
@@ -37,10 +43,15 @@ class ScanReportForm(forms.Form):
         queryset=Dataset.objects.order_by("name"),
         widget=forms.Select(attrs={"class": "form-control"}),
     )
+    visibility = forms.ChoiceField(
```
spco, line 46: One last comment: does this belong in this PR? Seems like it should be in #403 (which will be replacing it with React anyway)

api/mapping/services_rules.py
spco: I agree that we shouldn't have a bare except here. If nothing can go wrong, then remove the try/except. If something can go wrong, then handle each case specifically. Otherwise, any error in the above will silently be glossed over, which will be enormously difficult to debug.

api/mapping/forms.py
vpnu: Yeah I only needed it at the time for Scan Report upload to work, so I rebased on that branch, I'll remove it from this PR

api/mapping/services_rules.py
```diff
@@ -841,3 +842,125 @@ def remove_mapping_rules(request, scan_report_id):
...
+            descendants = ConceptAncestor.objects.filter(ancestor_concept_id=rule)
+            # Find the ancestors of that rule
+            ancestors = ConceptAncestor.objects.filter(descendant_concept_id=rule)
+            # For each descendant found
+            for descendant in descendants:
+                # get the concept id
+                desc = descendant.descendant_concept_id
+                # Check if the descendant is a mapped concept in any other scan report
+                # and that it's not the same as the original concept
+                if (desc in all_mapping_rules) & (desc != rule):
```
spco, line 916: please use `and`

api/mapping/services_rules.py
vpnu: Changed it to show this in the terminal if there are no matches in other Scan Reports:
"No ancestors or descendants of <conceptID> appear in any other Scan Reports"

Also added a message in the pop-up modal to say "No ancestors or descendants of these mappings appear in any other Scan Reports" instead of just being empty

api/mapping/services_rules.py
```diff
@@ -841,3 +842,141 @@ def remove_mapping_rules(request, scan_report_id):
...
+            else:
+                raise Exception(
+                    "No ancestors or descendants of {} appear in any other Scan Reports".format(
+                        rule_name
+                    )
+                )
+            # Reset lists before moving on to the next mapping rule
+            descendant_list = []
+            ancestors_list = []
+        except Exception as e:
```
spco, line 975: I think this will need changing, because `except Exception` will be triggered by other exceptions that inherit from `Exception`, like `KeyError` or `ArithmeticError`, so this is no better than a bare `except`. 

I think better would be to just add a `print` statement in the else clause, and remove the try-except. As written now, _any_ error would trigger the error message, which would be possibly misleading if it's really a `KeyError`or something else triggering it.

dcl10 (approved):

PR #395, opened 2022-02-21
Refactor mapping reuse code

spco:
Initially, this started out as
- Refactor the pagination code to enable a 2-dimensional pagination of two lists. This includes extracting the maximum characters for a GET request to a variable, removing the unused paginate_chars(), adding handle_max_chars() to reuse code, and changing the signature of paginate() as it no longer handles 2 lists.

However, the full PR goes further and applies some further refactoring and bug fixing.


spco:
mapping reuse - values

This shows the logic of the mapping reuse code for SRValues, for future reference!

PR #403, opened 2022-02-23
Upload scanreport react

Samad198:
Created the upload scanreport page in react
image
image
image
image

### DIFF
--- a/api/mapping/templates/mapping/upload_scan_report.html
+++ b/api/mapping/templates/mapping/upload_scan_report.html
@@ -26,7 +26,9 @@
                     data: formData,
                     enctype: 'multipart/form-data',
                 });
-                // need to test that it actually throws an error here if the upload is unsuccseeful
+                if (res.status == "form-invalid") {
+                    throw {statusText:"There was a form error",...res}
+                }
                 return res
             }
             catch (err) {
--- a/api/mapping/templates/mapping/upload_scan_report.html
+++ b/api/mapping/templates/mapping/upload_scan_report.html
@@ -9,67 +9,14 @@
             reports</a></li>
     </ol>
 </nav>
-<h2>New Scan Report</h2>
 
-<form method="post" name="simpleForm" id="simpleForm" action="#" enctype="multipart/form-data">{% csrf_token %}
+<div> 
+    <div id='root'></div>
+    <script type="text/javascript" >
+        var a = "{{a}}"
+        var u = "{{u}}"
+    </script>
+    <script type="module" src="../../../static/javascript/react/src/uploadScanReport.js"></script>
 
-    <div class="mb-3">
-        {{ form.parent_dataset.errors }}
-        <label class="form-label" id='parent_dataset'>{{ form.parent_dataset.label }}</label>
-        {{ form.parent_dataset }}
-    </div>
-    
-    <div class="mb-3">
-        {{ form.dataset.errors }}
-        <label class="form-label">{{ form.dataset.label}}</label>
-        {{ form.dataset }}
-    </div>
-
-    <div class="mb-3">
-
-        <label class="form-label">{{ form.scan_report_file.label }}</label>
-        {{ form.scan_report_file }}
-        {%if form.scan_report_file.errors%}
-        <div class="alert alert-danger" role="alert">{{ form.scan_report_file.errors }}</div>
-        {%endif%}
-
-    </div>
-
-    <div class="mb-3">
-
-        <label class="form-label">{{ form.data_dictionary_file.label }}</label>
-        {{ form.data_dictionary_file }}
-        {%if form.data_dictionary_file.errors%}
-        <div class="alert alert-danger" role="alert">{{ form.data_dictionary_file.errors }}</div>
-        {%endif%}
-
-    </div>
-    <div class="row mx-auto">
-        <div class="spinner-border mx-auto d-none" name="spinner" id="spinner" role="status">
-          <span class="sr-only">Uploading...</span>
-        </div>
-    </div>
-    <button type="submit" class="btn btn-primary" id="btn">Submit</button>
-
-</form>
-<script>
-
-$("#simpleForm").on('submit',function(ev) {      
-        $.ajax({
-        type: "POST",
-        data: {
-        'csrfmiddlewaretoken': '{{ csrf_token }}',
-        },
-        beforeSend: function (data) {
-        $('#spinner').addClass('d-none');
-        },
-        success: function (data) {
-            $('#spinner').removeClass('d-none');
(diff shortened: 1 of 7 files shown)


api/mapping/views.py
```diff
@@ -279,6 +283,7 @@ class DatasetFilterView(generics.ListAPIView):
     filter_backends = [DjangoFilterBackend]
     filterset_fields = {
         "id": ["in"],
+        "data_partner": ["in","exact"],
```
dcl10, line 286: Let's not do this change. `DatasetFilterView` gets merged into `DatasetListView` by #392. 
```suggestion
```

react-client-app/src/api/values.js
```diff
@@ -25,6 +25,10 @@ const usePost = async (url,data) =>{
         body: JSON.stringify(data)    
     }
     );
+    if (response.status < 200 || response.status > 300) {
+        console.log(response)
+        throw response
+    }
```
dcl10, line 31: What happens to cause the status to be less than 200 or greater than 300? Is this bit just to show you something in the terminal when something goes wrong?

api/mapping/urls.py
```diff
@@ -264,9 +269,9 @@
         name="password_change_done",
     ),
     # Project URLs
-    path("projects", views.ProjectListView.as_view(), name="project_list"),
+    path("api/projects/", views.ProjectListView.as_view(), name="project_list"),
     path(
-        "projects/<int:pk>",
+        "api/projects/<int:pk>/",
```
dcl10, line 274: Thank you.

api/mapping/templates/mapping/upload_scan_report.html
```diff
@@ -6,70 +6,37 @@
...
+                    enctype: 'multipart/form-data',
+                });
+                if (res.status == "form-invalid") {
+                    throw {statusText:"There was a form error",...res}
+                }
+                return res
+            }
+            catch (err) {
+                throw err
+            }
```
dcl10, line 36: In #404 I made it so that the CSRF token can be obtained from the cookies. This then means the `usePost` and `usePatch` function in `values.js` automatically get the CSRF token for you, and you should be able to send the form data using `usePost` no problem. 
We should merge #404's code into here and ditch this ajax call here in favour of `usePost`

api/mapping/views.py
```diff
@@ -968,6 +973,10 @@ class ScanReportFormView(FormView):
     template_name = "mapping/upload_scan_report.html"
     success_url = reverse_lazy("scan-report-list")
 
+    def form_invalid(self, form):
+        error_dict= {'status':'form-invalid','form-errors':form.errors}
+        return HttpResponse(json.dumps(error_dict),content_type="application/json")
+
```
dcl10, line 979: Is this the bit that gives back the errors for your screenshot on the PR?

react-client-app/src/components/UploadScanReport.jsx
```diff
@@ -0,0 +1,398 @@
...
+        setUsersList(usersQuery)
+    }, []);
+
+    useEffect(async () => {
+        setFormErrors({ ...formErrors, datapartner: undefined })
+        // change the dataset list
+        setLoadingDataset(true)
+        const dataPartnerId = selectedDataPartner.id
+        if (dataPartnerId != undefined) {
+            const datasets_query = await useGet(`/datasetsfilter/?data_partner=${dataPartnerId}`)
```
dcl10, line 56: after merging #392, you can change this to:
```suggestion
            const datasets_query = await useGet(`/datasets/?data_partner=${dataPartnerId}`)
```

react-client-app/src/components/UploadScanReport.jsx
```diff
@@ -0,0 +1,398 @@
...
+                visibility: projectVisibleToPublic ? "PUBLIC" : "RESTRICTED"
+            }
+            if (!projectVisibleToPublic) {
+                data.viewers = users.map(item => item.id)
+            }
+            const newDataset = await usePost(`/datasets/create/`, data)
+            await mapDatasetToProjects(newDataset, projects)
+            setLoadingDataset(true)
+            // revalidate
+            const datasets_query = await useGet(`/datasetsfilter/?data_partner=${data_partner.id}`)
```
dcl10, line 111: same here
```suggestion
            const datasets_query = await useGet(`/datasets/?data_partner=${data_partner.id}`)
```

dcl10 (changes requested):

react-client-app/src/api/values.js
Samad198: If the status is less than 200 or more than 300 then there was an error on the server, and the usePost function throws an error

api/mapping/views.py
Samad198: Yes it is

react-client-app/src/api/values.js
dcl10: Is it not greater than 500 for server errors? 400-499 is for when there a problem with the information the client is sending, if I'm not mistaken. Is this meant to simply catch when something goes wrong when doing `usePost`?

react-client-app/src/api/values.js
```diff
@@ -27,6 +27,21 @@ const usePost = async (url,data) =>{
     const res = await response.json();
     return res;
 }
+const postForm = async (url,data) =>{
+    const response = await fetch(`${url}`,
```
dcl10, line 31:
```suggestion
    const response = await fetch(url,
```

react-client-app/src/components/UploadScanReport.jsx
```diff
@@ -0,0 +1,398 @@
...
+            let formData = new FormData()
+            await formData.append('parent_dataset', selectedDataset.id)
+            await formData.append('dataset', scanReportName.current.value)
+            await formData.append('scan_report_file', whiteRabbitScanReport)
+            await formData.append('data_dictionary_file', dataDictionary)
+            setUploadLoading(true)
+
+            const response = await postForm(window.location.href,formData)
+            // redirect if the upload was successful, otherwise show the error message
+            window.location.href = `${window.u}scanreports/`
```
dcl10, line 183: `window.u` will not work anymore. I think if you want to change just the path you use `window.location.pathname` right?
```suggestion
            window.location.pathname = `/scanreports/`
```

dcl10 (changes requested):

react-client-app/src/api/values.js
Samad198: Yeah it's just to catch when there is any error status in usePost

dcl10 (dismissed):
LGTM. Just need to sort out the conflict in `views.py`

Issue #406, opened 2022-02-25
specimen not working


calmacx:
## Problem

* specimen concept:
 4001225 ..  | 119297000 | Blood specimen | Specimen | Standard [Blood specimen](https://athena.ohdsi.org/search-terms/terms/4001225)  
 ---> not working  

![image](https://user-images.githubusercontent.com/69473770/155710344-e603a941-0cb0-4d98-bdcb-4d8987d7c6e3.png)

## Suspected problem

* Throws an error when trying to grab the "source_concept_id" 
```
 const destination_field = await cachedOmopFunction(fields,domain+"_source_concept_id")
    // if a destination field can't be found for concept domain, return error                                                                                  
    if(destination_field == undefined){
        throw 'Could not find a destination field for this concept'
    }
```
https://www.ohdsi.org/web/wiki/doku.php?id=documentation:cdm:specimen

* specimen table does not have a `source_concept_id`

## Fix(es)

1) if statement to skip making a rule for the `specimen_source_concept_id` in the react code
2) why do we still bother saving the `<domain>_source_concept_id`? 
    --> If we force standard concept ids (which we do), then any `<domain>_source_concept_id == <domain>_concept_id ` 
    --> ... the purpose of source_concept_id in OMOP is if you want to use a Non-Standard concept ID for the source value and relate it to a Standard concept ID
   --> Does RQuest ever query the source_concept_id? I bet not.
   --> we could remove the rule creation for `<domain>_source_concept_id` and if we wanted we could get the ETL tool to automatically fill it from the  `<domain>_concept_id`

PR #407, opened 2022-02-25
Feature/add/specimen table

calmacx:
Temporary fix for #406


calmacx:
@spco @dcl10 this one is related to #406 that i've been able to test now that devel been fixed.

I'll keep it as a draft for now (not completely all tests), unless you want to go for it.

The issue is that the specimen table doesnt have a `specimen_source_concept_id` like the other tables, that I didn't notice.

It exposes the ad-hoc/hacky rule building that we do, but I think the catch I put in solves the problem.

Other point to note, that I've written in a few other places, is that I don't see why we bother with building rules for the `<domain>_source_concept_id` still.

This was a legacy from when I was coding in the case where you might want to have a Non-Standard concept or if the data already had a health data code associated to it (e.g ICD10).

No-one ever told me what RQuest does/uses. But I'd gamble it doesn't care and only performs searches on the concept_id. 

Cutting out a source_concept_id from the rules and therefore ETL could reduce the unnecessary doubling of processing the concept_id (because concept_id === source_concept_id)  ..and could always be done with the automatic filling feature in the ETL tool if really needed.

react-client-app/src/api/values.js
```diff
@@ -197,8 +199,8 @@ const saveMappingRules = async (scan_report_concept,scan_report_value,table) =>
         'drug_exposure':['drug_exposure_start_datetime','drug_exposure_end_datetime'],
 	'procedure_occurrence':['procedure_datetime'],
 	'specimen':['specimen_datetime']
-        }
-    const destination_field = await cachedOmopFunction(fields,domain+"_source_concept_id")
+        }      
+    const destination_field = await cachedOmopFunction(fields,domain+"_concept_id")
```
calmacx, line 203: as above, unsafe to assume domain+"_source_concept_id" exists for all omop tables... it does for concept_id.. i think

react-client-app/src/api/values.js
```diff
@@ -233,13 +235,23 @@ const saveMappingRules = async (scan_report_concept,scan_report_value,table) =>
...
+    //check if this table is in the list of tables that should be skipped (i.e. have no _source_concept_id)
+    //note - I see no need for keeping a rule for the source_concept_id for any table anymore
+    //     - https://github.com/CO-CONNECT/mapping-pipeline/issues/406
+    //     - Calum 25/02/2022
+    let tempOmopField; 
+    if(! m_skip_source_concept_id.includes(omopTable.table)){
+         tempOmopField = await cachedOmopFunction(fields,domain+"_source_concept_id")
+	 data.omop_field = tempOmopField.id
+	 promises.push(usePost(`/mappingrules/`,data))
+     }
```
calmacx, line 253: switch the code to do the source concept here instead of 

side note: we really dont need source concept anyway (need to still triple check with Phil/Tom)

react-client-app/src/components/MappingTbl.jsx
```diff
@@ -107,20 +107,88 @@ const MappingTbl = (props) => {
...
+					   await saveMappingRules(scan_report_concept,scan_report_value,scan_report_table);
+				     }
+				     //console.log(`done with loop ${i}`)
+			       });//end of promises
+			       await Promise.all(promises);
+			       //refresh when done
+			       window.location.reload(true);
+			 })
+		  .catch( (err) =>  console.log(err));
+      }
```
calmacx, line 189: an attempt to switch the looping over scan_report_concepts to the react code..

It is difficult as you need to retrieve various data for the scan_report_concept to find the associated scan_report_value and table that is needed by this `saveMappingRules` function

spco:
This is to be shelved for now, as we may not need to support the specimen table. I will close this PR for now to keep the place tidy, but we should remember it and resurrect it later if required.

I will open a separate PR to remove specimen table from the list of allowed tables, so that users cannot accidentally add specimen concepts that would then break the building of all rules in a ScanReport in w away that's hard to understand and debug.

PR #408, opened 2022-02-25
added name column to scanreport table

Samad198:
image


spco (commented):
All looks good, but could you please switch the order so it is ID, DataPartner, Dataset, Name.

PR #410, opened 2022-03-01
added uniqueness check to dataset names

Samad198:
Requires migration


vpnu (changes requested):
could you also add these model changes in the changelog?

spco (approved):

vpnu (approved):

PR #414, opened 2022-03-02
Remove 2 dimensional pagination and use `logging` module

spco:
This PR 

1.  Removes the 2-dimensional pagination code in favour of multiple calls to 1-dimensional pagination. This results in (a) fewer API calls; and (b) less code to maintain
2. Additionally removes all `print()` statements in `ProcessQueue` in favour of using the `logging` module. This defaults to only display the `INFO` level messages, but this can be changed to also show all `DEBUG` messages by simply modifying the `logger.setLevel` line.


spco:
Ready for review @vpnu

vpnu (approved):

PR #416, opened 2022-03-04
Feature/dataset admin form

dcl10:
# Changes
- Created django view to render an admin form at `/datasets/admin/<dataset id>`
- Created the form to allow admin users to update a datasets
  - If the user is not an admin the form will not allow them to perform the update
  - An error will show if the user tries to supply a blank name
 
# Screenshots
## User successfully updates the dataset
Screenshot 2022-03-04 at 15 54 59

## Form when visibility is set to RESTRICTED
Screenshot 2022-03-04 at 15 55 33

## Form when visibility is set to PUBLIC
Screenshot 2022-03-04 at 15 56 33

## User tries to supply no name
Screenshot 2022-03-04 at 15 54 00

## Non-admin tries to perform update
Screenshot 2022-03-04 at 16 01 38

### DIFF
--- a/react-client-app/src/components/DatasetAdminForm.jsx
+++ b/react-client-app/src/components/DatasetAdminForm.jsx
@@ -15,21 +15,18 @@ const DatasetAdminForm = ({ setTitle }) => {
     // Set up component state
     const [alert, setAlert] = useState({ hidden: true, title: '', description: '', status: 'error' })
     const [dataset, setDataset] = useState({})
+    const [visibility, setVisibility] = useState()
     const [loadingMessage, setLoadingMessage] = useState("Loading page")
     const [formErrors, setFormErrors] = useState({});
 
-    // Set up component refs
-    const datasetName = useRef()
-    const datasetVisibility = useRef()
-    const datasetDataPartner = useRef()
-    const datasetViewers = useRef()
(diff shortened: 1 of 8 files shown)


react-client-app/src/components/DatasetAdminForm.jsx
```diff
@@ -0,0 +1,278 @@
...
+
+    // Set up page
+    useEffect(
+        async () => {
+            setTitle(null)
+            // Get dataset
+            const datasetQuery = await useGet(`/datasets/${datasetId}`)
+            setDataset(datasetQuery)
+            setIsPublic(datasetQuery.visibility === "PUBLIC")
+            const dataPartnerQuery = await useGet("/datapartners/")
```
Samad198, line 55: Could speed this up by running them all in parallel in a promise and using Promise.all() to get the results, if the queries aren't reliant on eachother

spco:
LGTM once you async as @Samad198 suggests

react-client-app/src/components/DatasetAdminForm.jsx
dcl10: @Samad198 is that different to just putting `await` in front of lines 53 and 54?

vpnu (approved):

PR #420, opened 2022-03-08
Fix DataDictionary creation to match new model

vpnu:
DataDictionary foreign key to scan_report was removed from the model, so removed the link to scan_report in ScanReportFormView where we create a data dictionary


dcl10 (approved):

dcl10:
I'm happy if @spco is happy

PR #425, opened 2022-03-11
fixed mapping json download

Samad198:
# Changes

- Fixed broken downloading of mapping json and downloading of mapping csv caused by model changes

# Migrations

# Screenshots
- The pull request fixes the bug caused by the migration changes, but I have found that download mapping csv for 
/scanreports/40/mapping_rules/
still throws an unrelated error.
I believe it is thrown on this line inside services_rules.py with this value inside the content variable
image
image

# Checks

- [x] Run migrations, if any.
- [ ] Run unit tests.

### DIFF
--- a/api/mapping/services_rules.py
+++ b/api/mapping/services_rules.py
@@ -614,7 +614,7 @@ def get_mapping_rules_json(structural_mapping_rules):
     # add the metadata and cdm object together
     return {"metadata": metadata, "cdm": cdm}
 
- 
+
 def download_mapping_rules(request, qs):
     # get the mapping rules
     output = get_mapping_rules_json(qs)
--- a/api/mapping/services_rules.py
+++ b/api/mapping/services_rules.py
@@ -642,7 +642,7 @@ def download_mapping_rules_as_csv(request, qs):
     return_type = "csv"
     fname = f"{scan_report.parent_dataset.data_partner.name}_{scan_report.dataset}_structural_mapping.{return_type}"
     # return a response that downloads the csv file
-
+ 
     # make a string buffer
     _buffer = io.StringIO()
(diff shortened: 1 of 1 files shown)


calmacx:
@Samad198 are you saying you wanting me to look at what is going on in `services_rules.py` ? to figure out what is still causing a crash?

Samad198:
> @Samad198 are you saying you wanting me to look at what is going on in `services_rules.py` ? to figure out what is still causing a crash?

Yes please, i'm not really sure how the csv code works

Samad198:
Should have included that this is the error that's thrown
image

spco:
I think the issue here is that the value beginning  `SARS-COV-2` contains a comma and thus needs escaping. We have set `csv.QOUTE_NONE`, sand so it raises an `Error` as per https://docs.python.org/3/library/csv.html#csv.QUOTE_NONE .

I think in the `csv.writer` declaration you can supply `escapechar='\'`. It doesn't matter too much what escapechar we use as the CSV is only for human-readability. @calmacx do you see an issue if we do that?

spco:
Actually, perhaps I'd prefer is we didn't add `escapechar` and instead switched to using `QUOTE_MINIMAL` in place of `QUOTE_NONE`. Any objections?

calmacx:
No objections @spco  - though do we even need this csv download stuff? Was it suppose to be for NHS-D, who wanted to do their own mapping ?

spco:
As it's here I think we may as well keep it - it is not high-maintenance.

spco (changes requested):
Just replace QUOTE_NONE with QUOTE_MINIMAL

spco (approved):

PR #428, opened 2022-03-15
Dataset page

Samad198:
# Changes
- Edited the dataset admin page to show immutable values if the user is not an admin
- Added a link to the dataset admin page in the scanreport list page

# Migrations
# Screenshots

If the user is not an admin, they will see this
image
Otherwise, they will see
image

# Checks

- [ ] Run migrations, if any.
- [ ] Run unit tests.

### DIFF
--- a/api/mapping/templates/mapping/admin_dataset_form.html
+++ b/api/mapping/templates/mapping/admin_dataset_form.html
@@ -12,6 +12,9 @@
 
 <div>
     <div id='root'></div>
+    <script type="text/javascript" >
+        var currentUser = "{{current_user}}"
+    </script>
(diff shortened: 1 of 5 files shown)


spco (approved):
LGTM. I wonder whether we should change the URL as `datasets/admin` is a bit misleading when one isn't admin, but that is a separate concern from this PR.

spco:
Once this is merged, please double-check that everything is working on ccom-dev.

spco:
And just make sure the overlap with #429 works - slightly concerned they may both implement the functionality to test for admin status to allow edits to some fields.

dcl10 (commented):
LGTM. I'll try and merge this stuff with mine after this is merged.

PR #437, opened 2022-03-28
Dataset content

Samad198:
# Changes

- Added Dataset content page which displays list of scanreports in a dataset
- Changed link in scanreport page to go to this page

# Migrations

# Screenshots

image

#Notes 
- I'm thinking maybe it would be better to merge this PR first then refactor to make the table component reusable in a new PR?
# Checks

- [x] Run migrations, if any.
- [x] Run unit tests.


api/mapping/views.py
```diff
@@ -272,6 +272,8 @@ def get_serializer_class(self):
             # use the edit serialiser when the user tries to alter the scan report
             return ScanReportEditSerializer
         return super().get_serializer_class()
+    filter_backends = [DjangoFilterBackend]
+    filterset_fields = {"parent_dataset": ["exact"]}
```
dcl10, line 276: A small stylistic change. Could you put these 2 class variables at the top of the class please? I think it makes more sense to have them there like in the other views.

spco:
Please:

Remove the "New Scan Report" button.
Change title to "Scan Reports in <dataset name>". We don't need to show whether it's Active or Archived in the title, but keeping the Active/Archived buttons is useful.

Agreed - merge this one because it's in the "critical path" to the testing session, and then we can raise one for refactoring independently.

react-client-app/src/App.jsx
```diff
@@ -14,6 +14,7 @@ import ScanReportAdminForm from './views/ScanReportAdminForm'
 import Home from './components/Home';
 import { getScanReportConcepts, m_allowed_tables, useDelete, useGet, usePost, mapConceptToOmopField, saveMappingRules } from './api/values'
 import UploadScanReport from './components/UploadScanReport'
+import DatasetsContent from './components/DatasetsContent'
```
dcl10, line 17: I would put `DatasetsContent.jsx` in `./views` now with the scan report admin form view. This isn't really a "component" seeing as is essentially the whole page. The same is true for other "components" in the `components` folder. But we can change them later. For this PR, just move `DatasetsContent`.

dcl10 (changes requested):
In `DatasetContent.jsx` could you go through replace all the `==` and `!=` with `===` and `!==`, please. This checks equality of type and value and is much safer. I don't think we need the flexibility of `"true"` being equal to `true`, etc. Just a quick find and replace job in the IDE.

dcl10:
> Please:
> 
> Remove the "New Scan Report" button. Change title to "Scan Reports in ". We don't need to show whether it's Active or Archived in the title, but keeping the Active/Archived buttons is useful.
> 
> Agreed - merge this one because it's in the "critical path" to the testing session, and then we can raise one for refactoring independently.

I would also add the button to the details page of the scan report on the rows too.

Samad198:
image
The page now looks like this

Samad198:
> image
> 
> The page now looks like this

Just spotted the mistake in the table. Will fix now

dcl10:
> image
> 
> The page now looks like this

You need to add a blank header to the header row to fix that, @Samad198

Samad198:
image

Samad198:
image

dcl10 (approved):

spco:
LGTM - merge if you're happy @Samad198

PR #439, opened 2022-03-29
Bugfix/incomplete scan report query

dcl10:
# Changes

* Updated scan report list query to take into account editors, admins and authors
* Changed Scan Report upload procedure.
  - Scan Report inherits visibility from parent Dataset.

# Migrations
# Screenshots

* code is formatted correctly
Screenshot 2022-03-29 at 10 51 31

# Checks

- [X] Run migrations, if any.
- [X] Update `changelog.md`, including migration instructions if any.
- [X] Run unit tests.

### DIFF
--- a/api/mapping/views.py
+++ b/api/mapping/views.py
@@ -1224,9 +1224,6 @@ class ScanReportFormView(FormView):
         )
 
         scan_report.author = self.request.user
-        if parent_dataset.visibility == VisibilityChoices.RESTRICTED:
-            scan_report.viewers.set(parent_dataset.viewers.all())
-            scan_report.editors.set(parent_dataset.editors.all())
         scan_report.save()
 
         # Grab Azure storage credentials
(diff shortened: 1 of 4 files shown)


api/mapping/views.py
```diff
@@ -287,24 +287,99 @@ def get_queryset(self):
             return ScanReport.objects.all().distinct()
 
         return ScanReport.objects.filter(
+            # parent dataset and SR are public checks
             Q(
-                parent_dataset__project__members=self.request.user.id,
+                # parent dataset and SR are public
                 parent_dataset__visibility=VisibilityChoices.PUBLIC,
```
spco, line 293: Does this mean if the Dataset and SR are public, that it doesn't matter whether you're a member of a Project linked to the Dataset? Or is there a filter on Project membership before this?

dcl10: I moved that check down to the bottom of the query since it's the same for all cases. Because it acts like a `kwarg` while the big `Q` statement is an `arg`, it has to go last.

PR #434, opened 2022-03-24
Feature/scan report admin page

dcl10:
# Changes

* Created new reusable form components:
  - `CCTextInput`: for text inputs.
  - `CCSwitchInput`: for boolean switches.
  - `CCSelectInput`: for selecting a single choice.
  - `CCMultiSelectInput`: for selecting multiple choices.
* Added admin form for Scan Reports on frontend.
  - Found at `/scanreports/<scanreport_id>/admin`.
  - Scan report admins can change various attributes of a scan report.
  - *Cannot* change files associated with the scan report.

# Screenshots

- Successfully update a scan report.
Screenshot 2022-03-24 at 15 15 46

- User is not an admin
Screenshot 2022-03-24 at 16 45 06

# Checks

**Important:** please complete these **before** merging.
- [X] Run migrations, if any.
- [X] Update `changelog.md`, including migration instructions if any.
- [X] Run unit tests.

PR #440, opened 2022-03-29
Dataset list page

vpnu:
# Changes

- Added Dataset List Page
- Display ID, Name, Data Partner, Visibility and Creation Date. Allow filtering on Data Partner. 
- Link to list of Scan reports in a dataset 
- Add link to dataset detail page
 
# Migrations

# Screenshots
Screenshot 2022-03-29 at 16 27 02

# Checks

- [ ] Run migrations, if any.
- [ ] Run unit tests.


spco:
Please change the title from "Datasets" to "Dataset #<dataset_id>: <dataset name>". I'm happy otherwise.

api/mapping/templates/mapping/dataset_list.html
```diff
@@ -0,0 +1,20 @@
...
+        <li class="breadcrumb-item"><a href="{% url 'home' %}">Home</a></li>
+        <li class="breadcrumb-item"><a href="{% url 'datasets' %}">Datasets</a></li>
+    </ol>
+</nav>
+
+<div>
+    <div id='root'></div>
+    <script type="text/javascript" >
+        var isAdmin = "{{is_admin}}" === "True" ? true : false
+    </script>
```
dcl10, line 16: I don't think you don't need this since you're not passing `is_admin` from the `dataset_list_page` view.

api/templates/base.html
```diff
@@ -52,6 +52,9 @@
                 <li class="nav-item active">
                     <a class="nav-link" aria-current="page" href="{% url 'scan-report-list' %}">Scan Reports</a>
                 </li>
+                <li class="nav-item active">
+                    <a class="nav-link" aria-current="page" href="{% url 'datasets' %}">Datasets</a>
+                </li>
```
dcl10, line 57: Could we perhaps put the Datasets link before the Scan Reports one? Seeing how the Datasets contain all the Scan Reports

react-client-app/src/App.jsx
```diff
@@ -14,6 +14,7 @@ import ScanReportAdminForm from './views/ScanReportAdminForm'
 import Home from './components/Home';
 import { getScanReportConcepts, m_allowed_tables, useDelete, useGet, usePost, mapConceptToOmopField, saveMappingRules } from './api/values'
 import UploadScanReport from './components/UploadScanReport'
+import DatasetTbl from './components/DatasetTbl'
```
dcl10, line 17: Same as with #437 and #434, since this "component" is essentially a whole page and not a  reusable piece used on multiple different pages, could this go in `./views/` instead of `./components/` pls?

react-client-app/src/components/DatasetTbl.jsx
```diff
@@ -0,0 +1,138 @@
...
+    const [loadingMessage, setLoadingMessage] = useState("Loading Datasets")
+    const [datapartnerFilter, setDataPartnerFilter] = useState("All");
+    const [title, setTitle] = useState("Datasets");
+
+    useEffect(async () => {
+        // run on initial page load
+        props.setTitle(null);
+        // get datasets and sort by id
+        let datasets = await useGet(`/datasets/`);
+        datasets = datasets.sort((b, a) => (a.id > b.id) ? 1 : ((b.id > a.id) ? -1 : 0));
```
dcl10, line 19: I wonder if it's better to sort by creation date, recent to older, instead. I feel like users will want to see more recent entries first

react-client-app/src/components/DatasetTbl.jsx
```diff
@@ -0,0 +1,138 @@
...
+
+    if (loadingMessage) {
+        //Render Loading State
+        return (
+            <div>
+                <Flex padding="30px">
+                    <Spinner />
+                    <Flex marginLeft="10px">{loadingMessage}</Flex>
+                </Flex>
+            </div>
```
dcl10, line 74: We use this in tonnes of places. We should convert to a component in another PR

spco: Agreed, keep for another PR

react-client-app/src/components/DatasetTbl.jsx
vpnu: This is sorting by id, descending. The recently uploaded datasets are on top of the page so it does the same thing as sorting by creation date would

dcl10: Makes sense

PR #453, opened 2022-03-31
Dataset users

Samad198:
# Changes
- Made users only show users in the datasets project(s) that are active and members of 1 of the projects

# Migrations

# Screenshots
image

image

image

- Dataset inside co-connect
image

# Checks

- [ ] Run migrations, if any.
- [ ] Run unit tests.

### DIFF
--- a/api/mapping/views.py
+++ b/api/mapping/views.py
@@ -210,7 +210,6 @@ class ProjectListView(ListAPIView):
     """
 
     permission_classes = []
-    queryset = Project.objects.all()
     filter_backends = [DjangoFilterBackend]
     filterset_fields = {"name": ["in", "exact"]}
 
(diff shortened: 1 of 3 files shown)


api/mapping/views.py
```diff
@@ -222,6 +221,13 @@ def get_serializer_class(self):
             return ProjectSerializer
         return ProjectNameSerializer
 
+    def get_queryset(self):
+        if self.request.GET.get("dataset") != None:
+            dataset = self.request.GET.get("dataset")
+            return Project.objects.filter(datasets__exact=dataset).distinct()
```
dcl10, line 227:
```suggestion
        if dataset := self.request.GET.get("dataset"):
            return Project.objects.filter(datasets__exact=dataset).distinct()
```

dcl10: Not sure, but I suspect you can just do `...filter(datasets=dataset).distinct()`

spco:
LGTM, please just edit the text at the top of this page to show it's also checking project membership.

dcl10 (approved):

PR #459, opened 2022-04-08
Page refactors

Samad198:
# Changes
## Implemented the following views without django views
- Can now view scanreport tables for a given scanreport at /scanreports/XXXX
-  Can now view fields for a given table at scanreports/XXX/tables/YYY
- Can now view values for a given field at scanreports/XXX/tables/YYY/fields/ZZZ
- Can now edit table at scanreports/XXX/tables/YYY/update
- Can now edit field at )scanreports/ZZZ/tables/XXX/fields/YYY/update
## 404 page
- created 404 page for when resources can't be found

# Migrations
# Screenshots
- 404 page
Screenshot 2022-04-12 at 10 37 23

# Checks

- [ ] Run migrations, if any.
- [ ] Run unit tests.

### DIFF
--- a/api/mapping/views.py
+++ b/api/mapping/views.py
@@ -1889,25 +1889,25 @@ def scanreport_admin_page(request, pk):
     else:
         args["is_admin"] = False
     
-        return render(request, "mapping/admin_scanreport_form.html", args)
+    return render(request, "mapping/admin_scanreport_form.html", args)
 
 @login_required
-def scanreport_table_page(self,request, pk):
+def scanreport_table_page(request, pk):
     args = {}
     scan_report = None
     scan_report_name = None
     scan_report_table = None
     data_dictionary = None
-    
-    if len(self.get_queryset()) > 0:
-        scan_report = self.get_queryset()[0].scan_report
-        scan_report_name = scan_report.name
-        scan_report_table = self.get_queryset()[0]
-        try:
-            data_dictionary = scan_report.data_dictionary
-        except:
-            data_dictionary = None
-    
+
+    scan_report = ScanReport.objects.get(id=pk)
+    scan_report_name = scan_report.name
+    # not sure how to do this or if the variable is even used
+    #scan_report_table = self.get_queryset()[0]
+
+    try:
+        data_dictionary = scan_report.data_dictionary
+    except:
+        data_dictionary = None
 
     args["pk"] = pk
     args["scan_report"] = scan_report
--- a/api/mapping/views.py
+++ b/api/mapping/views.py
@@ -1978,3 +1978,24 @@ def scanreport_values_page(request, sr,tbl,pk):
     
     return render(request, "mapping/scanreportvalue_list.html", args)
 
+@login_required
+def update_scanreport_field_page(request, sr,tbl,pk):
+    args = {}
+    scan_report = None
+    scan_report_field = None
+    scan_report_table = None
+    scan_report_value = None
+
+    scan_report_value = ScanReportValue.objects.get(id=pk)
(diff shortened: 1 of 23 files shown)


api/mapping/views.py
```diff
@@ -1890,3 +1912,93 @@ def scanreport_admin_page(request, pk):
...
+
+@login_required
+def scanreport_table_page(request, pk):
+    args = {}
+    scan_report = None
+    scan_report_name = None
+    scan_report_table = None
+    data_dictionary = None
+
+    scan_report = ScanReport.objects.get(id=pk)
```
spco, line 1925: There are a couple of outstanding issues here (and they may be present at the values/fields level as well).

1. 
image

Here we're in SR 317, and the table `cdw.csv` links to the URL `http://localhost:8080/fields/?search=1248`. However, if I try to access that table through the new URL:
image
Having a bit of a play around, I seem to be able to access `317/tables/1` to `317/tables/417`, but no higher. We have over 5000 entries in the `mapping_scanreporttable` table. So something fishy is going on.

2. Relatedly, if I put in an SR ID and table ID that aren't even paired together, then I get a rather nonsensical result:
image
This is the table with global ID 1, so is unrelated to SR 317. But the page merrily loads, and the breadcrumbs along the top happily give the impression that this is a table from `EAVE-II-v2` when this is not the case. This should be returning 404 (or other if authentication gets in the way before that point).

spco (changes requested):

api/mapping/views.py
```diff
@@ -1845,7 +1845,9 @@ def list(self, request, pk):
         # scan_report = ScanReportSerializer(scan_reports, many=False).data
         # Set Storage Account connection string
         print(scan_report)
+        # TODO: `name` is not always defined, it seems
```
spco, line 1848: As I recall, we used to use the `file` field (`FileField`), but now we use `name` (`CharField`) to hold the filename. One more thing to add to the "must clean up" pile after 2.0 ;)

api/mapping/views.py
```diff
@@ -1890,3 +1914,71 @@ def scanreport_admin_page(request, pk):
...
+    args["pk"] = pk
+    args["scan_report"] = scan_report
+    args["scan_report_field"] = scan_report_field
+    args["scan_report_table"] = scan_report_table
+
+    return render(request, "mapping/scanreportvalue_list.html", args)
+
+
+@login_required
+def update_scanreport_field_page(request, sr, tbl, pk):
```
spco, line 1966: Is this still required?

spco (commented):
Looking good so far. When selecting an incorrect table from an SR (and field from table etc) we should 404, rather than this currently:
image

Could you also please explain where we're at with the xlsx download, and why? Thanks.

I note that the page e.g. `http://localhost:8080/tables/?search=317` is now quite broken - renders _all_ tables, and the link to Scan Report Details is broken. I presume the intention is to remove that view entirely, once we're happy the new view sufficiently replaces it?

api/mapping/views.py
dcl10: It looks like this is the method using the new URL patterns, so I think so

spco (commented):
All looking good so far, but access to tables/fields/values still needs to be restricted based upon the SR they belong to. So for example, I can't see SR 900 in the SR list, 
image
which is the correct behaviour, but I _can_ currently see `http://localhost:8080/scanreports/900` and `http://localhost:8080/scanreports/900/tables/7564`. I shouldn't be able to do see those, as I have only viewer permissions on the parent Dataset, and no permissions on the restricted SR 900.

PR #460, opened 2022-04-11
Allow user to Archive Datasets, add hidden field to Dataset model, di…

vpnu:
…splay in DatasetTbl view

# Changes
- Create hidden field in Dataset Model
- Add Archive/Active button to Datasets page
- Add Filter for active/archived datasets to Datasets page

# Migrations
- Add `hidden` `BooleanField` to `Dataset`.

# Screenshots
- Page with Active/Archive filters and buttons

Screenshot 2022-04-11 at 15 19 21

# Checks

- [x] Run migrations, if any.
- [ ] Run unit tests.

### DIFF
--- a/react-client-app/src/components/UploadScanReport.jsx
+++ b/react-client-app/src/components/UploadScanReport.jsx
@@ -53,7 +53,7 @@ const UploadScanReport = ({ setTitle }) => {
         setLoadingDataset(true)
         const dataPartnerId = selectedDataPartner.id
         if (dataPartnerId != undefined) {
-            const datasets_query = await useGet(`/datasets/?data_partner=${dataPartnerId}`)
+            const datasets_query = await useGet(`/datasets/?data_partner=${dataPartnerId}&hidden=false`)
             // if currently selected dataset is in the list of new datasets then leave selected datasets the same, otherwise, make dataset equal to null 
             setDatasets([{ name: "------" }, ...datasets_query.sort((a, b) => a.name.localeCompare(b.name))])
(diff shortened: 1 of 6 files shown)


dcl10 (changes requested):

dcl10 (approved):
Subject to @vpnu doing the migration on dev, LGTM

react-client-app/src/components/UploadScanReport.jsx
```diff
@@ -53,7 +53,7 @@ const UploadScanReport = ({ setTitle }) => {
         setLoadingDataset(true)
         const dataPartnerId = selectedDataPartner.id
         if (dataPartnerId != undefined) {
-            const datasets_query = await useGet(`/datasets/?data_partner=${dataPartnerId}`)
+            const datasets_query = await useGet(`/datasets/?data_partner=${dataPartnerId}&hidden=false`)
```
spco, line 56:
```suggestion
            // Show only Datasets that have this Data Partner and are no archived
            const datasets_query = await useGet(`/datasets/?data_partner=${dataPartnerId}&hidden=false`)
```
Just being a bit more explicit

spco (commented):
Looks good. What are your thoughts on 55825? I think we _should_ remove mappings from consideration if their Dataset is archived, in the same way that we do with archived SRs. That is then consistent and gives probably the expected behaviour when the user just wants to archive the whole Dataset rather than having to go through all its SRs, archiving each. Agree/disagree?

PR #471, opened 2022-04-25
added information option to reusable input

Samad198:
# Changes
- Added info option to reusable input that displays an info icon with some information in a tooltip
-  Fixed Flex wrap not applying in the reusable component

# Migrations

# Screenshots
image

# Checks

- [ ] Run migrations, if any.
- [ ] Run unit tests.

### DIFF
--- a/react-client-app/src/components/CCMultiSelectInput.jsx
+++ b/react-client-app/src/components/CCMultiSelectInput.jsx
@@ -1,8 +1,9 @@
 import React from 'react'
 import ConceptTag from './ConceptTag'
 import {
-    Flex, FormControl, FormLabel, FormErrorMessage, Select
+    Flex, FormControl, FormLabel, FormErrorMessage, Select, Tooltip,Box, Wrap, WrapItem
 } from "@chakra-ui/react"
+import { InfoIcon } from '@chakra-ui/icons'
 
 const CCMultiSelectInput = (props) => {
     /**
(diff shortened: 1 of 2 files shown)


dcl10:
It now looks like the label and tags list in `CCMultiSelectInput` cover up the border of the dropdown menu.

One thing that I think might look better: The tooltip could be placed before the labels of the inputs. That way they're the first bit the user sees and they stay in the same place - instead of being shifted along by the tags (as in `CCMultiSelectInput`.

Samad198:
Info icon now shows up on the left side like this
image

react-client-app/src/components/CCMultiSelectInput.jsx
```diff
@@ -41,23 +42,30 @@ const CCMultiSelectInput = (props) => {
 
     return (
         <FormControl isInvalid={props.formErrors && props.formErrors.length > 0} mt={4}>
-            <Flex flexWrap={true}>
```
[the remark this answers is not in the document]
Samad198: That flexwrap didn't seem to be working for some reason. I'll add a margine bottom to stop that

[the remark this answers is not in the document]
Samad198: Just added the margin and it now looks like this
image

react-client-app/src/components/UploadScanReport.jsx
```diff
@@ -491,6 +491,7 @@ const UploadScanReport = ({ setTitle }) => {
             </Box>
 
             <CCMultiSelectInput
+                info="Info about this goes here"
```
dcl10, line 494: Can we leave these out until we have some information to put there please?

react-client-app/src/components/UploadScanReport.jsx
```diff
@@ -501,6 +502,7 @@ const UploadScanReport = ({ setTitle }) => {
             />
 
             <CCMultiSelectInput
+                info="Info about this goes here"
```
dcl10, line 505: same here

Samad198: yeah i'll remove those now

Samad198: Done

dcl10 (approved):

PR #472, opened 2022-04-26
added check to stop duplicate concepts

Samad198:
# Changes
- Added check on backend to make sure user can't add multiple concepts to the same object

# Migrations

# Screenshots
- This is now shown when user tries to add a duplicate concept
image

# Checks

- [ ] Run migrations, if any.
- [ ] Run unit tests.

### DIFF
--- a/api/mapping/views.py
+++ b/api/mapping/views.py
@@ -599,7 +599,18 @@ class ScanReportConceptViewSet(viewsets.ModelViewSet):
         )
         if concept.count() > 0:
             print("Can't add multiple concepts of the same id to the same object")
-            return Response(status=status.HTTP_400_BAD_REQUEST)
+            response = JsonResponse(
+                {
+                    "status_code": 400,
+                    "ok": False,
+                    "statusText": "Can't add multiple concepts of the same id to the same object",
+                }
+            )
+            response.status_code = 400
+            response.statusText = (
+                "Can't add multiple concepts of the same id to the same object"
+            )
+            return response
 
         serializer = self.get_serializer(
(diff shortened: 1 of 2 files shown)


spco:
Where does `Can't add multiple concepts of the same id to the same object` get printed - just to the console? Is there any way to have that show in the alert box, so the user has more specific feedback?

Samad198:
> Where does `Can't add multiple concepts of the same id to the same object` get printed - just to the console? Is there any way to have that show in the alert box, so the user has more specific feedback?

Yeah there should be a way to get it to show in the alert box. I'll try implementing that now

Samad198:
Alert now looks like this when the error is thrown
image

api/mapping/views.py
```diff
@@ -591,6 +591,27 @@ class ScanReportConceptViewSet(viewsets.ModelViewSet):
...
+                {
+                    "status_code": 400,
+                    "ok": False,
+                    "statusText": "Can't add multiple concepts of the same id to the same object",
+                }
+            )
+            response.status_code = 400
+            response.statusText = (
+                "Can't add multiple concepts of the same id to the same object"
+            )
```
spco, line 612: Looks like this is repeated content that needs to only be set once?

Samad198: The content inside the object is what is returned in the body and what is set on line 609 determines the status code of the overall response. line 610 is not needed though and I have now removed it

spco: So just to be clear, line 609 is required even though 604 sets the status_code? They are different?

PR #501, opened 2022-05-04
Datasets admin projects endpoints

Samad198:
# Changes
- Added endpoint to count how many projects are on a dataset
- endpoint to get all projects on a specific dataset already existed
- Added check on if the editor can see all projects and warning on editing if they cannot

# Migrations

# Screenshots
image

# Notes 
- Need to check if the existing endpoint that gets all projects in a specific dataset or if it only gets what the user is permitted to see

# Checks

- [ ] Run migrations, if any.
- [ ] Run unit tests.

### DIFF
--- a/api/mapping/urls.py
+++ b/api/mapping/urls.py
@@ -124,7 +124,11 @@ routers.register(
 routers.register(r"analyse", views.AnalyseRules, basename="getanalysis")
 
 urlpatterns = [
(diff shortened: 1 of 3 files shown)


spco:
Question: why in this screenshot does it say it's in 3 other projects? Did you create 3 other projects beyond CO-CONNECT, and TestProject2 to put it in?

spco:
Ok thanks!

PR #506, opened 2022-05-06
Bugfix/error 500 when no mapping rules

dcl10:
# Changes

- Fixed error 500 when trying to build mapping rules diagram without rules.

# Migrations
# Screenshots
# Checks

- [ ] Run migrations, if any.
- [X] Update `changelog.md`, including migration instructions if any.
- [ ] Run unit tests.

### DIFF
--- a/api/mapping/services_rules.py
+++ b/api/mapping/services_rules.py
@@ -561,6 +561,10 @@ def get_mapping_rules_json(structural_mapping_rules):
         dict : formatted json that can be eaten by the TL-Tool
     """
 
+    # Return empty metadata and cdm if `structural_mapping_rules` is empty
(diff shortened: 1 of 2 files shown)


spco (changes requested):
I think I would prefer we fix the faulty logic in `get_mapping_rules_json()`. We have 
https://github.com/CO-CONNECT/mapping-pipeline/blob/4649804918453c096fe837a0fb9405e25d9e1d0f/api/mapping/services_rules.py#L566
which errors if `structural_mapping_rules` is empty. If it's empty, `get_mapping_rules_json()` should just `return {"metadata": {}, "cdm": {}}` which I hope will not break the later steps of processing.

PR #509, opened 2022-05-10
Project list page

vpnu:
# Changes

- Created Project List Page
- Lists Project names and links to Datasets in each project

# Screenshots

Screenshot 2022-05-10 at 10 26 00

# Checks

- [ ] Run migrations, if any.
- [ ] Run unit tests.

### DIFF
--- a/api/mapping/templates/mapping/project_list.html
+++ b/api/mapping/templates/mapping/project_list.html
@@ -1,0 +1,11 @@
+{% extends "base.html" %}
+
+{% block body %}
+
+<div>
+    <div id='root'></div>
+
+    <script type="module" src="/static/javascript/react/src/projectsTable.js"></script>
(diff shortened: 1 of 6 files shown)


vpnu:
Still to be done:

1. In the projects object, we only have the dataset id. We also need the dataset name for the link to the dataset to display the name. This is almost there, the Promise on line 33 works, but lines 35-42 don't actually replace the dataset id array with both the id and name from the datasetPromises. The code is trying to replace datasets:[id1,id2,...] (from the api response) with datasets:[{id:id1,name:"datasetname"},{id:id2,name:"datasetname2"}...] etc so we can save the name of each dataset.
2. Add a link to Projects page from nav bar
3. Test that only the projects/datasets the user has access to are displayed

PR #589, opened 2022-11-11
Update path to new npm version in entrypoint.sh.

spco:
# Checks

- [ ] Run migrations, if any.
- [ ] Run unit tests.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # Run npm build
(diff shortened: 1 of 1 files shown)


spco:
This is being closed as it's not clear how to actually fix all the dependencies. What I've learnt:

Node 16.18.3 is the LTS to use for npm 8+, which is required for use of the `overrides` keyword to set e.g. the minimum `tar version higher than the now un-supported `react-js-pagination` package sets.

But doing this gives annoying error messages during snowpack's `npm run build` phase, with no useful info to help solve the issue.

PR #598, opened 2023-11-13
Disable sr list stats

spco:
# Changes

- Disable the stats counting in the right-hand columns of the ScanReport list page. These were causing problems, and taking 45s+ to return results, destabilising the whole platform. For now, these just return "Disabled" text instead. The functionality was always a convenience function, and would be better implemented with regularly scheduled recalculation rather than on-the-fly computation.
- Rework the filter for the SR list API call. This is a little easier to read and (potentially) more stream-lined.

# Migrations

NA

# Screenshots

# Checks

- [x] Run migrations, if any.
- [ ] Run unit tests.

### DIFF
--- a/api/mapping/views.py
+++ b/api/mapping/views.py
@@ -302,129 +302,94 @@ class ScanReportListViewSet(viewsets.ModelViewSet):
         """
         If the User is the `AZ_FUNCTION_USER`, return all ScanReports.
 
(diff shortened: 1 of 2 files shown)


spco:
Before on dev system: 
image
image

After:
image

Counting stats is clearly fixed. Performance of the list query seems unaffected.

PR #602, opened 2023-11-30
Docs/135771 Add developer quickstart

AndyRae:
# Changes

Adds a developer quick start to get the app running, with corresponding documentation: https://github.com/HDRUK/CaRROT-Docs/pull/17

Includes:
- Adds a `docker-compose` to get the stack running.
- Updates the `sample-env` and `sample-local-settings` to use default values that work with the docker environment.
- Adds seed data fixtures to populate the `OmopTable` and `OmopField` tables.
- Adds python-dotenv as a dependency, to enable loading configuration from environment variables on all systems.
- Updates readme to point to the docs, it doesn't feel necessary to maintain both.

# Checks

- [ ] Run migrations, if any.
- [ ] Run unit tests.

### DIFF
--- a/sample-local.settings.json
+++ b/sample-local.settings.json
@@ -5,7 +5,7 @@
         "FUNCTIONS_WORKER_RUNTIME":"python",
         "STORAGE_CONN_STRING":"DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;QueueEndpoint=http://azurite:10001/devstoreaccount1;BlobEndpoint=http://azurite:10000/devstoreaccount1;",
         "NLP_API_KEY":"",
-        "APP_URL":"http://localhost:8080/",
+        "APP_URL":"http://localhost:8000/",
         "AZ_FUNCTION_KEY":"",
         "SCAN_REPORT_QUEUE_NAME":"scanreports-local",
         "NLP_QUEUE_NAME":"nlpqueue-local",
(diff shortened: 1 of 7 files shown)


sample-local.settings.json
```diff
@@ -1,11 +1,11 @@
 {
     "IsEncrypted":false,
     "Values": {
-        "AzureWebJobsStorage":"",
+        "AzureWebJobsStorage":"UseDevelopmentStorage=true",
         "FUNCTIONS_WORKER_RUNTIME":"python",
-        "STORAGE_CONN_STRING":"",
+        "STORAGE_CONN_STRING":"DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;QueueEndpoint=http://azurite:10001/devstoreaccount1;BlobEndpoint=http://azurite:10000/devstoreaccount1;",
```
dcl10, line 6: no secrets in the repo pls

AndyRae: Sorry wasn't clear - this isn't a secret as such, but it is the default connection string for Azurite: https://learn.microsoft.com/en-us/azure/storage/common/storage-use-emulator#authorize-with-shared-key-credentials.
Like I could clear it, and then point to those docs, but I feel a cleaner dev experience is to have it pre-existing - it's no different than having a connection string for a database.

dcl10: Oh right is that key the same no matter what then?

AndyRae: Yep it's the default to connect to the emulator, so just an default open secret really. See: https://github.com/search?q=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq&type=code

Jivanam: Can't we use `UseDevelopmentStorage=true`?

PR #638, opened 2024-02-19
635 Update repo structure

### LINKED ISSUE #635: Project structure / developer workflow
I think there are some small steps to be made to enable this repository to be easier to work with, including:

- [x] - Add a `/workers` directory to group the Azure functions together, including their dependencies. 
- [x] - Add `.vscode` workspaces to enable Python environments to be separate (functions and Django). Including `launch.json` files for the functions/webapp. 
- [x] - Use Poetry for dependencies, with versioning (currently very few of the dependencies are pinned to a version), and `pyproject.toml`. This repo is too big to use a simple `requirements.txt` (we'd get Poetry to output a `requirements.txt` in CI)
- [x] - Use `pre-commit` for linting, formatting, and typing - here's a simple config we used on Ai4Green: https://github.com/AI4Green/AI4Green/blob/main/.pre-commit-config.yaml can also include #543 
- [x] - Add a tests folder, and begin to use Pytest to support this. 

I'd be happy to make a PR to cover all of this, but open for discussion on any of the points, and any additional ideas.

AndyRae:
# Changes

Update the repository mostly for developer experience, including:
- Move the dependency management to use Poetry (we've been using this as standard in DRS Python products)
- Move application code to `app/` with:
    - `app/workers`: for Azure functions
    - `app/api`: for Django API
    - `app/react-client-frontend` for React Frontend
- Adds `pre-commit` with a config including black, iSort, flake8, and prettier
- Add `.vscode` workspaces, and launch configurations (VSCode dev experience is now just click run)
- Adds Poetry dependencies to the Docker, and Github actions - we're not using Poetry to manage the env, but just getting the `requirements.txt` out of it.

Does not add the test directory in #635  - I don't want to move out the existing Django tests just yet (so there would be nothing in there yet), but as we're implementing some tests generally, this will happen in a future PR.

Closes #635 

@spco If you're happy with these changes, I'll make a corresponding update to the documentation as well when it's approved. 

# Checks

- [x] Run migrations, if any.
- [x] Run unit tests.


pyproject.toml
```diff
@@ -0,0 +1,26 @@
+[tool.poetry]
+name = "carrot-mapper"
+version = "0.1.0"
+description = ""
+authors = ["John Smith <john@example.com>"]
```
AndyRae, line 5: `pyproject.toml` requires an `authors` field, so happy for suggestions on who/email this should be

spco: I don't believe we have a carrot email, so it's probably either me or Phil? I'm happy either way.

If you can add multiple, then both might be worth doing - I guess just need to check who's ok with their email being public here.

spco:
Does this fully address #639 too?

AndyRae:
> Does this fully address #639 too?

Nope that will come in a follow up PR, I don't want to change too much at once.

spco:
> Nope that will come in a follow up PR, I don't want to change too much at once.

image
What's that you say? 😆

spco (approved):

Issue #639, opened 2024-02-19
💡 Combine Github release workflows


AndyRae:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your proposal related to a problem or functionality gap?

Currently the release workflow for dev/test/production is 3 separate Github actions, working with Gitflow. So merging a new fix requires 3 separate PRs, artifacts, and deployments: https://github.com/Health-Informatics-UoN/CaRROT-Mapper/pulls?q=is%3Apr+is%3Aclosed

### Describe your proposal

I propose combining this into one workflow, utilising Github environments to deploy a single artifact from the main branch. An example here: https://github.com/Health-Informatics-UoN/rcc-monitor/actions/runs/7931015704

### I'm part of a Project Team

Yes - DRS

### Anything else?

It might not be done here - but I think it's worth also exploring when database migrations happen as part of CD, not sure if this currently a manual process?

### Are you willing to contribute to developing this feature?

✅ Yes, me or my team intend to do the development.

Issue #617, opened 2024-01-15
Automatic Rules Refresh on the Mapping Rules page


PhilAppleby:
Rules should be refreshed automatically on Mapping Rules page load.

Alternatively a warning should be displayed to alert the user of the need to refresh

spco:
Refresh is only required once, after all Person ID and Date Event columns are set in all the tables. Is a warning to that effect what's required?

Could combine it with automated refresh, e.g. if the rules have never been generated, attempt to do so on page load (with a placeholder message to explain the delay), and if the Person IDs and Date Events haven't been set then present a message to that effect?

erummas:
Hi Sam, 
There is a possibility that not all tables require mapping at that time and some may get mapped later. Delaying rule generation may not be the best idea. 
I think a warning message with an option just as the person goes to the 'rules page' may work stating that 'if use has added any new tables since the last mapping , rules need to be refreshed for generate new mappings'.
Another way could be that every time a table is edited it goes to the 'Update Table'  page with a button update "tablename.csv" so we refresh rules on that button click.

spco:
Remember that rules that are added manually after initial import do not require "Refresh rules" to be pressed - their rules are generated at the moment the Concept is added. We need only run "Refresh Rules" once, after all the Person ID and Date Event columns have been set. I would assume that that information can be immediately set in all cases, as these need to have been decided before upload to ensure the dataset is in the right form.

So the workflow is
- upload
- set all Person IDs and Date Events
- Refresh Rules (and never touch that button a second time)
- Do further manual mapping
- No need to run "Refresh Rules" again.

While adding the flexibility you suggest above might be nice, I think it would be a lot of work to support a workflow that we could reasonably avoid with minimal inconvenience to the user. They just need to have added Person ID and Date Event to all the tables before refreshing - we could add that logic into the warnings etc to ensure the refresh only ever runs once, at the point when it's appropriate.

erummas:
Got it, usually in cases of datasets when we have a lot of tables, it is hard to keep track of what tables we have mapped and completed mapping for and thus usually add these details only when we decide to map that table. May be we could think of another ways to cater this problem.
I think an error message on the 'Go to rules' button would suffice to let the user know that they need to select IDs and Event dates for all tables to be able to generate the first mapping. So once all that information is populated the button lets us navigate to the rules page.

spco:
[...]

Issue #612, opened 2024-01-15
Allow upload of large WhiteRabbit Scan reports


PhilAppleby:
The upload function times-out for large scan reports.

For example, for the UoD SHARE dataset - more detail will be supplied in comments

spco:
More definition/examples of "large" are required. #tables, #fields, #values, #reused mappings, or #vocab-mapped values. All of these should ideally be handled by CaRROT-Mapper, but examples are needing for testing.

AndyRae:
I think there are 3 things we can do here:

1. As discussed, separate out the creation of tables function from the processing, and processing is only triggered after the user has set their ID / date in the web app. I'm unsure how much time this will save, but it will help with making the code more robust, and make the UX workflow clearer. I've begun work on looking at this.
2. Use the database directly instead of the Django API - looking through, I don't think this would be much work. We would move the tables to a shared library in the code, and enable the functions to use it instead. 
3. Increase the timeout on the function - this depends on the plan.

AndyRae:
As discussed, we're continuing with our work to split out the Scan Report fields "upload" from the Concept "processing". 
Our first iteration of this to come soon will continue with the existing UI, but the Scan Report will be uploaded in the UI, with no processing taking place. We will then create Concepts per table (and get the reusable Concepts) when the user selects their "Person ID" and "Date Event" for the table, and selects: "Update Table".

After that, we can either put effort into:
1. Make this upload and processing faster, directly using the Django ORM rather than the REST API.
2. Design the UX into a multi step form to support uploading a Scan Report. Currently our outlined temporary UX means still having to traverse the UI to upload, select their table details and so on. Ideally we want this encapsulated into a supported form to guide the user through, with the option for #623 also. And then allow for viewing more detailed logs and status of the upload/process.

PR #643, opened 2024-02-28
Refactor Azure Functions

AndyRae:
# Changes

A refactor of the Scan Report upload / process in `app/workers`, that enables larger Scan Reports being uploaded. This is the first step as part of #612 , and unlocks work on #617 

The UX change is:
- User uploads their Scan Report.
- User sets the Person ID / Date Event for each table, the Concepts are now created and reuse of existing concepts.
- User clicks "Refresh Rules" to generate the mapping rules for the Scan Report. 

This behaviour is a significant change, and so the now "legacy" behaviour has been preserved, and the two can be switched between by `UPLOAD_ONLY` environment variable. This can be removed, and the `ProcessQueue` function, when we are satisfied the new behaviour is suitably robust.

Includes:
- A new `UploadQueue` Azure function, responsible for uploading a Scan Report.
- A new `CreateConcepts` Azure function, responsible for creating Scan Report Concepts, AND reusing existing ones.
- Whilst the existing logic is retained in these functions, they have been refactored from the original `ProcessQueue`, adding docstrings, types, move to some modularity, and general clean up.
- Add tests (Pytest) for `app/workers` (a test CI workflow will follow as part of #639 , including #424 )
- Added `AZ_FUNCTION_USER` to the `sample-env.txt`
- Updated `ScanReportTableFilterViewSet` to return a 403 if the user is not the Azure function user.

A corresponding PR will be attached for the documentation updates.

# Environment changes

This does necessitate changes to any deployed instances of CaRROT Mapper, including:
- Add Azure queues:
    - [ ] Upload queue
    - [ ] Create concepts queue

- Add environment variables to the Web app:
    - [ ] `UPLOAD_QUEUE_NAME`: Name of the Azure upload queue
    - [ ] `CREATE_CONCEPTS_QUEUE_NAME`: Name of the Azure queue
    - [ ] `UPLOAD_ONLY`: true

# Checks

- [x] Run migrations, if any.
- [x] Run unit tests.


app/api/mapping/views.py
```diff
@@ -1,82 +1,32 @@
+import base64
```
AndyRae, line 1: All this importing is now `iSort` pre-commit fixing things :)

app/api/mapping/views.py
```diff
@@ -720,34 +754,34 @@ class ScanReportActiveConceptFilterViewSet(viewsets.ModelViewSet):
...
-                    scan_report_field__scan_report_table__scan_report__parent_dataset__hidden=False,
-                    scan_report_field__scan_report_table__scan_report__status="COMPLET",
-                )
-                qs = ScanReportConcept.objects.filter(
-                    content_type=17, object_id__in=value_ids
-                )
-                return qs
-            return None
+        if self.request.user.username != os.getenv("AZ_FUNCTION_USER"):
+            raise PermissionDenied(
```
AndyRae, line 758: Refactors to use a guard instead, but raises a 403 for clarity

spco (approved):
While I can't claim to have checked every line, this looks good. Merging should probably be coordinated so that we can ensure all the environment changes are made at the same time etc.

https://carrot4omop.ac.uk/AzureFunctions/ProcessQueue/#processing will need updating, but I'd wait until the full process has been rewritten rather than making iterative changes to that now.

PR #655, opened 2024-03-20
Merge dev to prod 2.0.14

AndyRae:
Follow #645 to deploy new pipelines.

Relevant to #639


spco:
Is there a reason for skipping the `test` branch (we normally do `devel` -> `test` -> `master`), or is it purposeful, given the deployment changes?

AndyRae:
> Is there a reason for skipping the `test` branch (we normally do `devel` -> `test` -> `master`), or is it purposeful, given the deployment changes?

There's no benefit to it now really - as we're essentially making `devel` and `test` redundant as branches, as they don't have workflows that deploy from them anymore.

Once this is merged into `master`, it will deploy to `dev`, and have an approval process to deploy to `test`, and then to `production`. 

But I now realise, that there could be code that has been merged into `dev` -> `test` but not into `master` yet, so we would be skipping it in this case. I'll double check we're not leaving anything behind.

spco (approved):
Once you're happy with the state vs test, carry on.

PR #658, opened 2024-03-26
Fix ProcessQueue magic numbers

AndyRae:
# Changes

In #652 this updated the web API (amongst other things) to no longer use the hardcoded ids 15 / 17. 
But this broke the `ProcessQueue` function as it still uses them with the API. Although I'm expecting to delete this function soon, I'm keen to preserve it as a fallback option temporarily.

So this PR fixes the `ProcessQueue` to use the correct strings for the API.

# Checks

- [x] Run migrations, if any.
- [x] Run unit tests.

### DIFF
--- a/app/workers/ProcessQueue/__init__.py
+++ b/app/workers/ProcessQueue/__init__.py
@@ -1,18 +1,17 @@
+import asyncio
 import json
 import logging
 import os
-
 from collections import defaultdict
 from datetime import datetime
 
-import asyncio
+import azure.functions as func
 import httpx
 import requests
-import azure.functions as func
-
 from requests.models import HTTPError
 from shared_code import omop_helpers
-from . import helpers, blob_parser
+
+from . import blob_parser, helpers
 
 # import memory_profiler
 # root_logger = logging.getLogger()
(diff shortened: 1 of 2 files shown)


spco (approved):

app/workers/ProcessQueue/__init__.py
```diff
@@ -152,13 +151,13 @@ def select_concepts_to_post(
     for new_content_detail in new_content_details:
         try:
             # fields
-            if content_type == 15:
+            if content_type == "scanreportfield":
```
dcl10, line 154: For these hardcoded strings that you're reusing all over the place, can you create a constants file, like `constants.py` in the appropriate place to store these. Then import them and you only need to change in one place

app/workers/ProcessQueue/__init__.py
```diff
@@ -152,13 +151,13 @@ def select_concepts_to_post(
...
                 (
                     existing_content_id,
                     concept_id,
                 ) = details_to_id_and_concept_id_map[str(new_content_detail["name"])]
             # values
-            elif content_type == 17:
+            elif content_type == "scanreportvalue":
```
dcl10, line 160: Same idea here and all the other ones

PR #675, opened 2024-04-11
Add serverless mapping rules

### LINKED ISSUE #612: Allow upload of large WhiteRabbit Scan reports

### LINKED ISSUE #663: Mapping Rules Azure Function
# Summary

Add an Azure function to generate mapping rules.

Make Mapping Rules happen on a table basis.
Take a list of Scan Reports that should be reused. If there is no Scan Reports passed, just use them all. (#623) (I'm not sure, but assume this probably has to happen earlier in the mapping process)

Relevant to: #612 

We will retain the existing "Refresh Rules" button for now, as a fallback option. 

# Tasks

- [x] New Azure Function
- [x] Generate mapping rules (port existing code)
- [x] Add the trigger from re use of mapping rules

# Acceptance Criteria

- [x] Mapping rules are generated with Azure functions
- [x] Tests pass

AndyRae:
# Changes

Adds a new Azure Function `MappingRules`, that generates mapping rules when given a `table_id`.

The UX change is that reused mapping rules will now be generated when a user selects the Person Id and Date Event for that table, instead of using the "Refresh Rules" button.
This enables larger scan reports, as the workload is now serverless, and only happening on a table basis rather than a whole Scan Report. 

The UI has been updated with a notice to reflect this, and removing the "Refresh Rules" button entirely. The JSX changes are mostly Prettier working away.

The code to generate the rules is mostly lift and shift (added docs, types, tests), as we can now use the database directly in Azure functions.
I've moved this code into the shared package, with the intention of removing it from the CLI (which will use the new service), and the existing `services_rules.py`.

Closes #663 
Closes #612 

# Screenshots
Screenshot 2024-04-12 at 11 38 49

# Deployment

Needs the following environment variables added:
- [x] workers dev
- [x] workers test
- [x] workers prod

```
COCONNECT_DB_ENGINE
COCONNECT_DB_HOST
COCONNECT_DB_NAME
COCONNECT_DB_PASSWORD
COCONNECT_DB_PORT
COCONNECT_DB_USER
MAPPING_RULES_QUEUE_NAME
```
A new `MappingRules` queue for:
- [x] storage dev
- [x] storage test
- [x] storage prod

# Checks

- [x] Run migrations, if any.
- [x] Run unit tests.

### DIFF
--- a/app/react-client-app/src/components/EditTable.jsx
+++ b/app/react-client-app/src/components/EditTable.jsx
@@ -1,130 +1,175 @@
-import React, { useState, useEffect, useRef } from 'react'
-import { Button, Flex, Link, Spinner, useDisclosure, ScaleFade } from "@chakra-ui/react"
-import CCBreadcrumbBar from './CCBreadcrumbBar'
-import CCSelectInput from './CCSelectInput'
-import ToastAlert from '../components/ToastAlert'
-import PageHeading from './PageHeading'
-import { useGet, usePatch } from '../api/values'
+import React, { useState, useEffect, useRef } from "react";
+import {
+  Alert,
+  AlertIcon,
+  Button,
+  Flex,
+  Link,
+  Spinner,
+  useDisclosure,
+  ScaleFade,
+} from "@chakra-ui/react";
+import CCBreadcrumbBar from "./CCBreadcrumbBar";
+import CCSelectInput from "./CCSelectInput";
+import ToastAlert from "../components/ToastAlert";
+import PageHeading from "./PageHeading";
+import { useGet, usePatch } from "../api/values";
 const EditTable = (props) => {
-    const { isOpen, onOpen, onClose } = useDisclosure()
-    const [alert, setAlert] = useState({ hidden: true, title: '', description: '', status: 'error' })
-    const value = window.pk ? window.pk : window.location.href.split("tables/")[1].split("/")[0]
-    const [fields, setFields] = useState(null);
-    const [table, setTable] = useState(null);
-    const [selectedPerson, setPerson] = useState("------");
-    const [selectedDate, setDate] = useState("------");
-    const [loadingMessage, setLoadingMessage] = useState(null)
-    const canEdit = window.canEdit
-    const scanReport = useRef([]);
+  const { isOpen, onOpen, onClose } = useDisclosure();
+  const [alert, setAlert] = useState({
+    hidden: true,
+    title: "",
+    description: "",
+    status: "error",
+  });
+  const value = window.pk
+    ? window.pk
+    : window.location.href.split("tables/")[1].split("/")[0];
+  const [fields, setFields] = useState(null);
+  const [table, setTable] = useState(null);
+  const [selectedPerson, setPerson] = useState("------");
+  const [selectedDate, setDate] = useState("------");
+  const [loadingMessage, setLoadingMessage] = useState(null);
+  const canEdit = window.canEdit;
+  const scanReport = useRef([]);
 
-    useEffect(async () => {
-        props.setTitle(null)
-        // get scan report table to use to get tables 
-        const scanreporttable = await useGet(`/scanreporttables/${value}/`)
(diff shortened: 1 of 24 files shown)


Dev-Akashili (approved):

app/workers/mapping/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""
```
spco, line 1: Bit of a niggly one, but could I suggest we rename this project? `mapping` is not all that descriptive of what the `MappingRules` function does, and the name `mapping` is already used as the base Django app.

Or is that the whole point, and it's important to reuse the same name to pick up the base app?

changelog.md
```diff
@@ -7,6 +7,7 @@ Please append a line to the changelog for each change made.
 
 ### Improvements
 - Added a `shared` library for the database models, enabling both the Django and Azure Functions to directly access the database.
+- Created new Azure Function MappingRules, which generates the reusable rules for a Scan Report table, when the CreateConcepts has finished.
```
spco, line 10:
```suggestion
- Created new Azure Function MappingRules, which generates the mapping rules (from reuse or from vocabularies) for a Scan Report table, when the CreateConcepts has finished.
```

app/react-client-app/src/components/EditTable.jsx
```diff
@@ -1,130 +1,175 @@
...
+            {loadingMessage ? loadingMessage : "Loading page"}
+          </Flex>
+        </Flex>
+      </div>
     );
-}
+  }
+  return (
+    <div>
+      <CCBreadcrumbBar>
```
spco, line 113: One for the future, but it just reminds me we should at some point scrub the `CC` prefix (which was `coconnect`) to something more appropriate.

app/shared/test/services/settings.py
```diff
@@ -0,0 +1,28 @@
...
+INSTALLED_APPS = [
+    "django.contrib.admin",
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
+    "shared",
+]
+
+DATABASES = {
+    "default": {
+        "ENGINE": os.environ.get("COCONNECT_DB_ENGINE"),
```
spco, line 18: Same comment as above - we should rename the `COCONNECT_*` to `CARROT_*` or similar. To worry about later!

spco (dismissed):

app/react-client-app/src/components/EditTable.jsx
```diff
@@ -1,130 +1,175 @@
...
+            title={alert.title}
+            status={alert.status}
+            description={alert.description}
+          />
+        </ScaleFade>
+      )}
+      <Flex paddingTop={5} paddingBottom={5}>
+        <Alert status="info">
+          <AlertIcon />
+          Mapping Rules cannot be generated without the Person ID and Date Event
```
spco, line 144: I think this could be made clearer. Perhaps 
```
Mapping Rules cannot be generated without the Person ID and Date Event being set.
Once you set these, Mapping Rules will be generated for all Concepts currently associated to the table.
```

blargly (approved):

app/workers/mapping/__init__.py
AndyRae: Yes it is the whole point - as we now share the database models, they have a dependency on a `mapping` app existing in the Django project.

app/shared/test/services/settings.py
AndyRae: Agreed - one for the backlog

app/workers/mapping/__init__.py
spco: Gotcha!

app/react-client-app/src/components/EditTable.jsx
AndyRae: Have applied this change

blargly (approved):

spco (approved):

Issue #664, opened 2024-04-05
Backend technical debt


AndyRae:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your proposal related to a problem or functionality gap?

In development we've identified a few issues that would make the backend much easier to develop, maintain, and faster for users. 

Here are some ideas - I'm very open to opinions on any of them, and always looking for more as well. 

### Describe your proposal

- [x] #685
    - [x] `ScanReportListView` 
    - [x] `StructuralMappingTableListView`
    - [x] (double check for any more!)
(Views that get data by the API, but are duplicating calls, as the rendered data is all driven by the API. Remove the code that does this)

- [x] #684
    - [x] rename the app/api/api project to app/api/core or config (as it being API is just confusing) 
    - [x] create a new `app/api/api` app, move all `api` relevant code in there. 
    - [x] Retain the `mapping` app for Django views, and general mapping application logic.
(Using Django projects to provide this separation)

- [x] #699 
    - [x] Turn off the `ProcessQueue`, delete the function
    - [x] Delete temp code to switch between the two by `UPLOAD_ONLY` environment variable.
- [ ] #726
- [x] Delete commented out code, document functions, add types, remove unused code. 
- [x] Add `MyPy` type checker gradually
- [ ] A new "version" endpoint
- [x] #741
    - [x] Remove production from the existing CI/CD
    - [x] Remove test from behind a user gate. 
    - [x] Enable deploy to dev run from `workflow_dispatch` so we can deploy any branch there for deployed testing.
    - [x] Add a Next.js build check on raise PR

### I'm part of a Project Team

Yes - DRS

### Are you willing to contribute to developing this feature?

✅ Yes, me or my team intend to do the development.

PR #686, opened 2024-04-23
Restructure Django app

### LINKED ISSUE #684: Separate the Django Views from API
- [x]  rename the app/api/api project to app/api/core or config (as it being API is just confusing)
- [x]  create a new app/api/api app, move all api relevant code in there.
- [x]  Retain the mapping app for Django views, and general mapping application logic.
- [x] Fix proxy url 
- [x] Add the v2 endpoint

(Using Django projects to provide this separation)

AndyRae:
# Changes

Restructures the Django app to make it easier to work with, particularly for development to understand what is API driven, and what is Django views.

- Rename the `app/api/api` project to `config` (as it being API is just confusing)
- Create a new `app/api/api` app, move all api relevant code in there. (enables us to drop the prepend `api/` for every url, and do it just in `api/config`
- Retain the mapping app for Django views, and general mapping application logic.
- Fix proxying url regex to work only for that page
- Add the V2 `ScanReportListView` endpoint (with pagination, filtering etc.)

Closes #684 
Tracked by #664 

# Checks

- [x] Run migrations, if any.
- [x] Run unit tests.


app/api/api/routers.py
```diff
@@ -0,0 +1,122 @@
+from api import views
+from rest_framework import routers
+
+routers = routers.DefaultRouter()
```
dcl10, line 4: ⛏️ can we rename `routers` to `router` seeing for the class is `DefaultRouter`?

app/api/config/urls.py
```diff
@@ -0,0 +1,19 @@
...
+from django.urls import include, path
+
+urlpatterns = [
+    (
+        path("", include("proxy.urls"))
+        if os.environ.get("ENABLE_PROXY", "False").lower() == "true"
+        else None
+    ),
+    path("api/", include("api.urls")),
+    path("", include("mapping.urls")),
```
dcl10, line 13: I'm a little sceptical of two paths at `""`. What's happening here?

AndyRae: So the first one is the proxy - it's allowing us to override any urls if it's enabled, effectively instead of routing to the Django views, we can forward on to the Next app instead. It's a separate app to make this clear, rather than bury it in `mapping/urls.py`
The second one is the default, that's including the Django views.

app/api/api/routers.py
AndyRae: Yep good point - have pushed this change

dcl10 (approved):

blargly (approved):

PR #693, opened 2024-04-29
Add datasets list index page

### LINKED ISSUE #683: Datasets Index
Example: `/datasets/`

Frontend Tasks:

- [x] Recreate the existing UI
- [x] Define types in Typescript for what is needed
- [x] Add pagination
- [x] Add filtering
- [x] Add sorting
- [ ] Review

Backend Tasks:

- [x] New API endpoint
- [x] With pagination, filtering, and sorting
- [x] Testing 
- [x] Add the URLS to proxy it 
- [ ] Review

AndrewThien:
# Changes

This update adds the UI of Datasets List index page and changes the following Backend:

- Defined the columns and data types for Data table rendering Datasets list data
- Added API endpoints to get Datasets list and to archive/unarchive the selected dataset
- Added Proxy URLs (in app/api/proxy/urls.py)
- Implemented the server-side pagination, sorting and filtering
- Modify the Backend functions to enable server-side Sorting and Filtering in the table
- Modify actions of table rows
- Config the "Carrot" colour palette in Tailwind 

Close #683 

# Screenshots

- Server-side Pagination
Screenshot 2024-04-30 at 14 34 19

- Archive/Unarchive Function
Screenshot 2024-04-30 at 14 34 48

- Server-side Sorting
Screenshot 2024-04-30 at 14 35 12

- Server-side Filtering
Screenshot 2024-04-30 at 14 35 39

- Actions Menu
Screenshot 2024-04-30 at 14 38 28

# Checks

- [x] Run migrations, if any.
- [x] Run unit tests.


spco:
In the first 2 images, dataset no.2 is showing in both views. When first navigating to the page, we should show only Active Datasets (i.e. apply `hidden=False` by default), with the Active Datasets toggle enabled as shown.

AndrewThien:
> In the first 2 images, dataset no.2 is showing in both views. When first navigating to the page, we should show only Active Datasets (i.e. apply `hidden=False` by default), with the Active Datasets toggle enabled as shown.

Hi @spco ,
By default, the table will render the dataset having "hidden=fault". 
The pictures in the last comment were used incorrectly. I already modified the pictures in the latest comment. Thanks for it, Sam.

spco (approved):
From the user perspective this looks fine - I'm not one to inspect the JS.

Could you perhaps show an expanded Actions menu too please?

AndrewThien:
Here is the one, @spco 
Screenshot 2024-04-29 at 12 14 55

app/next-client-app/app/datasets/page.tsx
```diff
@@ -0,0 +1,85 @@
...
+            <BreadcrumbItem>
+              <BreadcrumbLink href="/datasets">Datasets</BreadcrumbLink>
+            </BreadcrumbItem>
+          </BreadcrumbList>
+        </Breadcrumb>
+      </div>
+      <div className="flex justify-between mt-3">
+        <h1 className="text-4xl font-semibold">Dataset List</h1>
+        <Link href="/">
+          <Button size="lg" className="text-md bg-[#475da7]">
```
AndyRae, line 42: As discussed - would be great to drive these colours from tailwind themeing, instead of specifying it every time. 

https://tailwindcss.com/docs/customizing-colors#using-custom-colors

But set them on the default variant inside `Button.tsx`, so we just do it once.
There might be a couple of other places this is used, so fixing them in this PR would be great. (Just search for all instances of `#475da7`)

app/next-client-app/components/ui/checkbox.tsx
```diff
@@ -0,0 +1,30 @@
+"use client"
```
AndyRae, line 1: As discussed - I don't think this component is used? If it isn't, just delete it (and then npm package)

app/next-client-app/next.config.js
```diff
@@ -3,7 +3,7 @@ const nextConfig = {
   output: "standalone",
   experimental: {
     serverActions: {
-      allowedOrigins: [process.env.BACKEND_URL],
+      allowedOrigins: [process.env.BACKEND_ORIGIN],
```
AndyRae, line 6: Do leave a comment as to what works for this value in local development - and I agree, that having it distinct from `BACKEND_URL` is helpful, as it might need to be set to something different when deployed.

app/next-client-app/types/index.ts
```diff
@@ -13,3 +13,27 @@ interface ScanReport {
   previous: string | null;
   results: ScanReportResult[];
 }
+
+interface DataSet {
```
AndyRae, line 17: This is really a `DataSetPage` I think.

app/next-client-app/types/index.ts
```diff
@@ -13,3 +13,27 @@ interface ScanReport {
...
+  count: number;
+  next: string | null;
+  previous: string | null;
+  results: DataSetResult[];
+}
+
+interface DataSetResult {
```
AndyRae, line 24: And I think this is then called a `DataSet`

app/next-client-app/types/index.ts
```diff
@@ -13,3 +13,27 @@ interface ScanReport {
...
+  id: number;
+  name: string;
+  hidden: boolean;
+  visibility: string;
+  created_at: string;
+  data_partner: DataPartnerData;
+  admins: string[];
+}
+
+interface DataPartnerData {
```
AndyRae, line 34: And this can be just called a `DataPartner`


a nice read on naming things: https://journal.stuffwithstuff.com/2009/06/05/naming-things-in-code/
s/o @beforan

AndyRae (changes requested):

app/next-client-app/next.config.js
AndyRae: As discussed - keep you original change here to `BACKEND_ORIGIN`, but just leave a comment here as to what works locally for you, and should be in our `.env` files.

app/next-client-app/app/datasets/page.tsx
AndyRae: The correct link is actually: https://tailwindcss.com/docs/customizing-colors#adding-additional-colors
so inside `tailwind.config.js`:
```
module.exports = {
  theme: {
    extend: {
      colors: {
        carrot: {
          ....
        },
      }
    },
```

AndyRae (approved):

Dev-Akashili (approved):

spco (dismissed):
Can't quite tell from the screenshots. So long as the default sort by timestamp, descending, is also applied to the archived view (it's not in the screenshot but that might not be accurate) as well as the unarchived, then LGTM.

AndyRae (approved):

Dev-Akashili (approved):

PR #728, opened 2024-05-29
Implementation of NextJS to Dataset - Scanreports list view

### LINKED ISSUE #723: Datasets detail - list of scan reports

AndrewThien:
# Changes

This PR adds the NextJS view of the Dataset - Scanreports list 

Closes #723 

# Screenshots

Screenshot 2024-05-29 at 11 10 59

# Checks

- [x] Run migrations, if any.
- [x] Run unit tests.


app/next-client-app/app/datasets/[id]/page.tsx
```diff
@@ -0,0 +1,104 @@
...
+              <BreadcrumbLink href="/datasets">Datasets</BreadcrumbLink>
+            </BreadcrumbItem>
+            <BreadcrumbSeparator>/</BreadcrumbSeparator>
+            <BreadcrumbItem>
+              <BreadcrumbLink href={`/datasets/${id}/`}>
+                {datasetName.name}
+              </BreadcrumbLink>
+            </BreadcrumbItem>
+          </BreadcrumbList>
+        </Breadcrumb>
```
Comment (Jivanam) [inline app/next-client-app/app/datasets/[id]/page.tsx:59]: suggestion: may be a separate component, where we simply pass the target URL and display name.

Jivanam (approved):

app/next-client-app/types/dataset.ts
```diff
@@ -21,3 +14,16 @@ interface DataPartner {
   created_at: string;
   updated_at: string;
 }
+
+interface DataSetSRList {
```
AndyRae, line 18: Just trying to get my head around this - why is this needed on top of `DataSet`? In `api/datasets.ts` they both are used as return types, but looking at the API, they both return this `DataSetSRList` object (just one in a list, one on it's own).

app/next-client-app/app/datasets/[id]/page.tsx
Comment (AndrewThien) [reply app/next-client-app/app/datasets/[id]/page.tsx:59]: this is a good idea. But it is also kind of complicated (and needs long props) to re-use the component in other, deeper pages (Scanreports Fields/Values, for example) cause it will need different id, tableids, ... and names for the links. If you have any methods to handle this, please let me know. Cheers!

app/next-client-app/types/dataset.ts
AndyRae: Actually I understand why! Forgot the dataset list endpoint isn't actually useful.

AndyRae (approved):

app/next-client-app/app/datasets/[id]/page.tsx
Comment (Jivanam) [reply app/next-client-app/app/datasets/[id]/page.tsx:59]: No worries. You could try creating a list containing display names and their corresponding URLs, and then passing this list as a prop. But, this may add overhead due to the need to create and manage the list.

Comment (AndyRae) [reply app/next-client-app/app/datasets/[id]/page.tsx:59]: Absolutely - this is a bit of technical debt I've been willing to accept, as I feel we _could_ spend a day (or so) working out the perfect breadcrumb component. We'll add it to the backlog

PR #735, opened 2024-06-03
Implementation of Users' scanreports permissions on UI rendering 

### LINKED ISSUE #733: Frontend permissions implemented
When a user views:

- Scan Report Tables
- Scan Report Fields
- Scan Report Values

Their access to the form fields are based on their permissions - this does not currently work on the Nextjs app.

There is a fetch to `getScanReportPermissions` which returns the users permissions for a scan report. This needs to be implemented on those pages.

- [ ] Get the permissions on the page as above
- [ ] Add the permissions data to the data that we pass to the tables, this is then used inside `columns.tsx` to then render/disable buttons based on permissions.

AndrewThien:
# Changes

This PR implemented the users' permissions on a scan report to render the according UI. It also makes the rendering of the Alert box in the Update table page conditionally.  

Closes #733 

# Screenshots
Screenshot 2024-06-03 at 11 52 07
Screenshot 2024-06-03 at 11 52 44

# Checks

- [x] Run migrations, if any.
- [x] Run unit tests.

### DIFF
--- a/app/next-client-app/app/scanreports/[id]/tables/[tableId]/fields/[fieldId]/page.tsx
+++ b/app/next-client-app/app/scanreports/[id]/tables/[tableId]/fields/[fieldId]/page.tsx
@@ -105,8 +105,8 @@ export default async function ScanReportsValue({
         <h1 className="text-4xl font-semibold">Values</h1>
       </div>
       <ButtonsRow
-        scanreportId={id}
-        tableId={tableId}
+        scanreportId={parseInt(id)}
+        tableId={parseInt(tableId)}
         permissions={permissions.permissions}
       />
       <div>
--- a/app/next-client-app/app/scanreports/[id]/tables/[tableId]/page.tsx
+++ b/app/next-client-app/app/scanreports/[id]/tables/[tableId]/page.tsx
@@ -95,8 +95,8 @@ export default async function ScanReportsField({
         <h1 className="text-4xl font-semibold">Fields</h1>
       </div>
       <ButtonsRow
-        scanreportId={id}
-        tableId={tableId}
+        scanreportId={parseInt(id)}
+        tableId={parseInt(tableId)}
         permissions={permissions.permissions}
       />
       <div>


MoizIqbal97 (approved):

app/next-client-app/api/scanreports.ts
```diff
@@ -48,7 +48,9 @@ export async function getScanReports(
  * @param id The Id of the Scan Report
  * @returns A object with a list of the users permissions.
  */
-export async function getScanReportPermissions(id: number): Promise<{}> {
+export async function getScanReportPermissions(
+  id: string
+): Promise<{ permissions: string[] }> {
```
Jivanam, line 53: Don't we have permissions type?

AndrewThien: Yes, I was using the wrong one. Already fixed this. Thanks for posting it!

MoizIqbal97 (dismissed):

Jivanam (dismissed):

app/next-client-app/components/scanreports/ButtonsRow.tsx
```diff
@@ -5,9 +5,11 @@ import { BookText, ChevronRight, Pencil } from "lucide-react";
 export function ButtonsRow({
   scanreportId,
   tableId,
+  permissions,
 }: {
   scanreportId: string;
   tableId: string;
+  permissions: string[];
```
AndyRae, line 12: Should this be of type `Permission[]`?

https://github.com/Health-Informatics-UoN/CaRROT-Mapper/blob/8d9b83936c438415caabc1073306261c50e57fdb/app/next-client-app/types/permissions.ts

app/next-client-app/components/update-SR-table/UpdateForm.tsx
```diff
@@ -17,9 +17,11 @@ export interface ShortFields {
 export function UpdateForm({
   scanreportFields,
   scanreportTable,
+  permissions,
 }: {
   scanreportFields: ShortFields[];
   scanreportTable: ScanReportTable;
+  permissions: string[];
```
AndyRae, line 24: As before - should this be of type `Permission[]`

https://github.com/Health-Informatics-UoN/CaRROT-Mapper/blob/8d9b83936c438415caabc1073306261c50e57fdb/app/next-client-app/types/permissions.ts

app/next-client-app/lib/concept-utils.ts
```diff
@@ -1,10 +1,11 @@
...
+// Method to add concepts and permissions to a scanreport result
 export const addConceptsToResults = <
-  T extends { id: number; concepts?: Concept[] }
+  T extends { id: number; concepts?: Concept[]; permissions: string[] }
 >(
   scanReportsResult: T[],
   scanReportsConcepts: ScanReportConcept[],
-  concepts: Concept[]
+  concepts: Concept[],
+  permissions: { permissions: string[] }
```
AndyRae, line 8: As before - should this be of type `Permission[]`?

https://github.com/Health-Informatics-UoN/CaRROT-Mapper/blob/8d9b83936c438415caabc1073306261c50e57fdb/app/next-client-app/types/permissions.ts

app/next-client-app/types/scanreport.ts
```diff
@@ -14,8 +14,9 @@ interface ScanReportTable {
   updated_at: Date;
   name: string;
   scan_report: number;
-  person_id: string;
-  date_event: string;
+  person_id: string | null;
+  date_event: string | null;
+  permissions: string[];
```
AndyRae, line 19: All the permissions here - as before, should be of the type `Permission[]`

https://github.com/Health-Informatics-UoN/CaRROT-Mapper/blob/8d9b83936c438415caabc1073306261c50e57fdb/app/next-client-app/types/permissions.ts

app/next-client-app/components/scanreports/ButtonsRow.tsx
```diff
@@ -27,11 +29,19 @@ export function ButtonsRow({
       </div>
       <div className="flex gap-2">
         {" "}
-        <Link href={`/scanreports/${scanreportId}/tables/${tableId}/update`}>
-          <Button variant="outline">
-            Edit Table <Pencil className="ml-2 size-4" />
+        {permissions.includes("CanEdit") || permissions.includes("CanAdmin") ? (
```
AndyRae, line 32: I can see why we need to do this - to find a way to not include the Link in the component based on the permissions. It's frustrating we can't disable the Link
I wonder if a cleaner way might be to extract the Button component itself.

```
const canEdit = permissions.includes("CanEdit") || permissions.includes("CanAdmin");

const EditTableButton = (
  <Button variant="outline" disabled={!canEdit}>
    Edit Table
    <Pencil className="ml-2 size-4" />
  </Button>
);

return canEditOrAdmin ? (
  <Link href={`/scanreports/${scanreportId}/tables/${tableId}/update`}>
    {EditTableButton}
  </Link>
) : (
  EditTableButton
);
```

At least that way we're not duplicating the Button code, still not clean though.

Looking at the Shadcn Button docs, the advised way to use a Link inside a Button is to wrap as following instead:
```
<Button asChild>
  <Link href="/login">Login</Link>
</Button>
```
I'm curious - if the button is disabled in this case, does the link no longer work?

AndyRae: Also if we do have to do it the first way, as we use this same pattern 3 times at least already - I think it's worth moving the entire logic out to a `EditButtonLink` component (or something similarly named)

Jivanam (approved):

AndyRae (approved):

PR #753, opened 2024-06-10
Delete scanreport backend

### LINKED ISSUE #729: Delete Scan Report backend
## Backend

Add the backend to allow a user to delete a Scan Report.

# Tasks
- [ ] Add a delete method to the scanreport endpoint.
- [ ] Add delete logic for the data. 
- [ ] Add deleting the files in blob storage, both the scan report, and the data dictionary. Consider doing this by a "service" module.
- [ ] Add permissions for only an Owner of a scan report to delete it.

Dev-Akashili:
# Changes
Closes #729 
- Added delete method to the scanreport v2 endpoint.
- Added method to delete scan report and data dictionary files in blob storage.

# Checks

- [x] Run migrations, if any.
- [x] Run unit tests.

### DIFF
--- a/app/api/api/views.py
+++ b/app/api/api/views.py
@@ -59,6 +59,7 @@ from mapping.permissions import (
     CanViewProject,
     get_user_permissions_on_scan_report,
 )
+from mapping.services import delete_blob
 from mapping.services_rules import get_mapping_rules_list
 from rest_framework import generics, status, viewsets
 from rest_framework.filters import OrderingFilter
@@ -390,7 +391,32 @@ class ScanReportListViewSetV2(ScanReportListViewSet):
     def get_serializer_class(self):
         if self.request.method in ["GET", "POST"]:
             return ScanReportViewSerializerV2
+        if self.request.method in ["DELETE"]:
+            return ScanReportTableEditSerializer
         return super().get_serializer_class()
+
+    def destroy(self, request, pk):
+        instance = self.get_object()
+        required_permissions = ["CanView", "CanEdit", "CanAdmin"]
+        permissions = get_user_permissions_on_scan_report(request, pk)
+        if not all(permission in permissions for permission in required_permissions):
+            raise PermissionDenied
+        try:
+            delete_blob(instance.name, "scanreports")
+        except Exception as e:
+            return Response(
+                {"detail": f"Error deleting scanreport: {e}"},
+                status=status.HTTP_404_NOT_FOUND,
+            )
+        try:
+            delete_blob(instance.data_dictionary, "data-dictionaries")
+        except Exception as e:
+            return Response(
+                {"detail": f"Error deleting data dictionary: {e}"},
+                status=status.HTTP_404_NOT_FOUND,
+            )
+        instance.delete()
+        return Response(status=status.HTTP_204_NO_CONTENT)
 
 
 class DatasetListView(generics.ListAPIView):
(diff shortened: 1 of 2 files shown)


app/api/mapping/services.py
```diff
@@ -46,3 +47,17 @@ def download_data_dictionary_blob(blob_name, container="data-dictionaries"):
     response["Content-Disposition"] = f'attachment; filename="{blob_name}"'
 
     return response
+
+
+def delete_blob(blob_name, container):
```
AndyRae, line 52: Please add type annotations (both the arguments, and return type) and a docstring, we're setting a new standard for the codebase, and any code introduced should use them.

app/api/mapping/services.py
```diff
@@ -46,3 +47,17 @@ def download_data_dictionary_blob(blob_name, container="data-dictionaries"):
...
+    try:
+        blob_service_client = BlobServiceClient.from_connection_string(
+            os.environ.get("STORAGE_CONN_STRING")
+        )
+        container_client = blob_service_client.get_container_client(container)
+        blob_dict_client = container_client.get_blob_client(blob_name)
+        # Delete the blob
+        blob_dict_client.delete_blob()
+        return True
+    except Exception as e:
```
AndyRae, line 62: I think it's fine to simplify this method, and not try/catch exceptions here. Especially as you're returning it anyway.

The service can just delete the blob - if it throws an error, the code calling it should handle it appropriately.

app/api/api/views.py
```diff
@@ -390,8 +391,33 @@ class ScanReportListViewSetV2(ScanReportListViewSet):
...
         if self.request.method in ["GET", "POST"]:
             return ScanReportViewSerializerV2
+        if self.request.method in ["DELETE"]:
+            return ScanReportTableEditSerializer
         return super().get_serializer_class()
 
+    def destroy(self, request, pk):
+        instance = self.get_object()
+        required_permissions = ["CanView", "CanEdit", "CanAdmin"]
+        permissions = get_user_permissions_on_scan_report(request, pk)
```
AndyRae, line 401: I hadn't realised, but there's no need for you to handle permissions. This is done on the base class, line 245. So you can delete any permission related code here :)

AndyRae (changes requested):

app/api/api/views.py
```diff
@@ -390,8 +391,33 @@ class ScanReportListViewSetV2(ScanReportListViewSet):
...
+        if not all(permission in permissions for permission in required_permissions):
+            raise PermissionDenied
+        try:
+            delete_blob(instance.name, "scanreports")
+        except Exception as e:
+            return Response(
+                {"detail": f"Error deleting scanreport: {e}"},
```
AndyRae, line 408: We wouldn't want to serialise the internal error to the user.

AndyRae (changes requested):

app/api/api/views.py
```diff
@@ -390,8 +391,24 @@ class ScanReportListViewSetV2(ScanReportListViewSet):
     def get_serializer_class(self):
         if self.request.method in ["GET", "POST"]:
             return ScanReportViewSerializerV2
+        if self.request.method in ["DELETE"]:
+            return ScanReportEditSerializer
         return super().get_serializer_class()
 
+    def perform_destroy(self, pk):
```
AndyRae, line 398: `perform_destroy` takes an instance
https://www.cdrf.co/3.3/rest_framework.mixins/DestroyModelMixin.html

AndyRae (changes requested):

PR #762, opened 2024-06-20
Fix Mapper stops responding when add/delete concept quickly 

### LINKED ISSUE #761: Mapper isn't responding

- [X] I have searched the existing issues

### Current Behavior

Hi, this has happened twice today that the mapper has stopped responding. At first I thought it may be due to some server issues but I think it happens when the concepts are added manually too quickly (I could be wrong) as both times it happened when I was entering the same concept_id for multiple codes so copy pasting quickly across various code values. Both times I am unable to refresh the application. 

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
- Other environment details:
```

_No response_

### Anything else?

_No response_

None

AndrewThien:
# Changes

This PR fixes the bug which is causing the Mapper to stop responding when adding/deleting concepts quickly.

Carrot now can handle deleting or adding concepts by the users who are doing them quickly, by using React.useState to update only the affected records, not revalidating the whole table after the actions, and using React.memo to prevent unnecessary tags loading.

Many many thanks and credits to @AndyRae 

Closes #761 

# Checks

- [x] Run migrations, if any.
- [x] Run unit tests.


spco:
It would be helpful if you could provide a paragraph here on what the issue was, and what the fix is? For posterity!

AndrewThien:
> posterity

Hi @spco, the fix and the original issue (#761) have been updated on the PR's comment. Thanks!

spco:
I was hoping for an explanation at a more technical level: what was the cause of the issue, rather than the symptom? And what (at a high level) is the fix? Thanks!

AndrewThien:
Hi @spco,
So the cause of the issue is that the page would re-fetch the whole table and its record's data (which is often heavy), instead of the data of the affected record, after a successful adding or deleting concept. This process took time (and was unnecessary) and while this process happened, if the users added/deleted other concepts, the app would stop responding/crash.
Therefore, the fix is making the app re-fetch only the data of the affected record, basically. --> Making the app respond faster and more effectively --> If users want to add/delete quickly concepts, they can do it now --> Fix the bug
Hope this helps :D

spco:
Ok, so just so I understand, is it technically a race condition? And if so, the fix is not to remove a race condition from possibly occurring, but just reduce the amount of work needed so the race condition is less likely to occur?

MoizIqbal97 (approved):

AndyRae (approved):

spco:
Thanks @AndrewThien

PR #777, opened 2024-07-01
Adding dataset bug fix

AndrewThien:
# Changes

This PR adds a fix for the Adding Dataset. 
When modifying the API for Updating the Dataset details, changes there accidentally made the API for creating a new Data set break. This PR fixed it now by adding a V2 Serialiser. This will be combined in the future with the Adding New Dataset API, potentially.

# Checks

- [x] Run migrations, if any.
- [x] Run unit tests.

### DIFF
--- a/app/api/api/views.py
+++ b/app/api/api/views.py
@@ -18,6 +18,7 @@ from api.serializers import (
     DatasetAndDataPartnerViewSerializer,
     DatasetEditSerializer,
     DatasetViewSerializer,
+    DatasetViewSerializerV2,
     DomainSerializer,
     DrugStrengthSerializer,
     GetRulesAnalysis,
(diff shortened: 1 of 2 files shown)


app/api/api/views.py
```diff
@@ -616,7 +617,7 @@ def get_queryset(self):
 
 
 class DatasetCreateView(generics.CreateAPIView):
-    serializer_class = DatasetViewSerializer
```
AndyRae, line 619: Could we do this the other way around? 

As in this view - the "legacy" form should still use `DatasetViewSerializer`

The form you had to update the serializer for should use `DatasetViewSerializerV2`

`V2` should be anything we upgrade

AndyRae (changes requested):

Jivanam (approved):

Jivanam (approved):

AndyRae (approved):

PR #779, opened 2024-07-01
Add Scan Report permissions

AndyRae:
# Changes

Adds a permissions filter to Scan Reports, tables, fields, and values views. 
These permissions are based on the existing in `permissions.py`

Related #776 

# Checks

- [x] Run migrations, if any.
- [x] Run unit tests.

### DIFF
--- a/app/api/api/filters.py
+++ b/app/api/api/filters.py
@@ -1,0 +1,145 @@
+from django.db.models.query_utils import Q
+from rest_framework import filters
+from shared.data.models import VisibilityChoices
+
+
+class CanAccessScanReportFilterBackend(filters.BaseFilterBackend):
+    """
+    Filter that only allows users to see Scan Reports they are allowed to view, edit, or admin.
+
+    """
+
+    # Each model type needs a different relationship to get the Scan Report / Dataset permissions.
+    RELATIONSHIP_MAPPING = {
+        "scanreport": "",
+        "scanreporttable": "scan_report__",
+        "scanreportfield": "scan_report_table__scan_report__",
+        "scanreportvalue": "scan_report_field__scan_report_table__scan_report__",
+    }
+
+    def filter_queryset(self, request, queryset, view):
+        """
+        Filters the queryset based on visibility and permissions related to datasets and scan reports.
+
+        Args:
+            - request: The request object.
+            - queryset: The queryset to be filtered based on visibility and permissions.
+
+        Returns:
+            - A filtered queryset based on the specified visibility and permission conditions.
+        """
+
+        model = queryset.model.__name__.lower()
+        relationship = self.RELATIONSHIP_MAPPING.get(model, "")
+
+        dataset_visibility = f"{relationship}parent_dataset__visibility"
+        scan_report_visibility = f"{relationship}visibility"
+        scan_report_viewers = f"{relationship}viewers"
+        scan_report_editors = f"{relationship}editors"
+        scan_report_author = f"{relationship}author"
+        dataset_viewers = f"{relationship}parent_dataset__viewers"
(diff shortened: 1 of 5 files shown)


dcl10 (approved):

app/api/api/views.py
```diff
@@ -1005,53 +1006,6 @@ class ScanReportConceptFilterViewSet(viewsets.ModelViewSet):
     }
 
 
-class ScanReportActiveConceptFilterViewSet(viewsets.ModelViewSet):
```
spco, line 1008: Has this been superceded already? This is/was the function used by the auto-mapping functions to look across all Active SRs for reusable mappings.

app/api/api/views.py
```diff
@@ -1266,18 +1220,6 @@ def get_serializer_class(self):
         return super().get_serializer_class()
 
 
-class ScanReportFilterViewSet(viewsets.ModelViewSet):
```
spco, line 1269: Have these all been superceded elsewhere?

app/api/api/filters.py
```diff
@@ -0,0 +1,145 @@
...
+        dataset_visibility = f"{relationship}parent_dataset__visibility"
+        scan_report_visibility = f"{relationship}visibility"
+        scan_report_viewers = f"{relationship}viewers"
+        scan_report_editors = f"{relationship}editors"
+        scan_report_author = f"{relationship}author"
+        dataset_viewers = f"{relationship}parent_dataset__viewers"
+        dataset_editors = f"{relationship}parent_dataset__editors"
+        dataset_admins = f"{relationship}parent_dataset__admins"
+
+        return queryset.filter(
```
spco, line 44: I haven't looked at the code for too long to be sure this isn't implemented elsewhere: but I think here we also need to implement Project-level filtering. Users should never be able to see a SR that isn't in one of the Projects they are a member of, even if that SR and its Dataset are set as "PUBLIC".

The only exception is AZ_FUNCTION_USER who can see everything to allow mapping reuse.

app/api/api/views.py
AndyRae: Yes the mapping reuse is completely database driven now - no API is used.

app/api/api/views.py
AndyRae: Yes they are no longer in use.

app/api/api/filters.py
AndyRae: Thanks - hadn't thought of the project level, I will add this.

dcl10: good spot @spco

dcl10 (approved):

AndrewThien (approved):

app/api/api/filters.py
```diff
@@ -0,0 +1,148 @@
...
+
+        dataset_visibility = f"{relationship}parent_dataset__visibility"
+        scan_report_visibility = f"{relationship}visibility"
+        scan_report_viewers = f"{relationship}viewers"
+        scan_report_editors = f"{relationship}editors"
+        scan_report_author = f"{relationship}author"
+        dataset_viewers = f"{relationship}parent_dataset__viewers"
+        dataset_editors = f"{relationship}parent_dataset__editors"
+        dataset_admins = f"{relationship}parent_dataset__admins"
+        project_members = f"{relationship}parent_dataset__project__members"
```
spco, line 43: I'm going to assume that you've checked this works, remembering that datasets to projects is many-to-many.

spco (approved):

app/api/api/filters.py
AndyRae: Yep it works, and with multiple users involved (as in, multiple users being in the lists) - I was expecting to use `__in` (or however Django does it) but it wasn't necessary.

PR #792, opened 2024-07-11
Implementation of NextJS to Scan Report Details Page

### LINKED ISSUE #749: Scan Report Details (edit form)
# Summary

Rewrite the Scan Report Details page in Nextjs, for example: https://ccom-dev.azurewebsites.net/scanreports/2565/details/

# Tasks

- [ ] Implement with FormIk
- [ ] Implement the React Select from `New Scan Report` form
- [ ] POST data - same pattern as Add Scan Report table.
- [ ] On form load - populate with the existing data
- [ ] Permissions for a Scan Report - CanAdmin | CanEdit
- [ ] Add information tooltips (this can mirror the `New Scan Report` form
- [ ] Testing

# Acceptance Criteria

- [ ] A working Scan Report Details form, with data posting, and populating.

AndrewThien:
# Changes
This PR implemented NextJS to build the Scan Report Details view. It also added 
- The IsAuthor checking class in permissions.py to check if the user is the author of the current scan report or not.
- Some tooltips for Save buttons in other forms to clarify the permissions
- Add a notice to add a new Dataset which has to have a unique name which will relate to finding Scan report details
- Make the data sent to the PATCH endpoint of updating Scan Report to be an abstract object

Closes #749 

# Screenshots

- General view
Screenshot 2024-07-11 at 12 23 39

# Checks

- [x] Run migrations, if any.
- [x] Run unit tests.

### DIFF
--- a/app/api/mapping/permissions.py
+++ b/app/api/mapping/permissions.py
@@ -39,6 +39,34 @@ def is_az_function_user(user: User) -> bool:
     return user.username == os.getenv("AZ_FUNCTION_USER")
 
 
+def is_scan_report_author(obj: Any, request: Request) -> bool:
+    """Check if the user is the author of a scan report.
+
+    Args:
+        obj (Any): The object to check (should be a ScanReport or related object).
+        request (Request): The request with the User instance.
+
+    Returns:
+        bool: `True` if the request's user is the author of the scan report, else `False`.
+    """
+    try:
+        # If `obj` is a scan report table|field|value, get the scan report
+        # it belongs to and check the user has permission to view it.
+        if sub_scan_report_query := SCAN_REPORT_QUERIES.get(type(obj)):
+            sub_scan_report = sub_scan_report_query(obj)
+        else:
+            sub_scan_report = obj
+
+        # Check if the user is the author
+        if isinstance(sub_scan_report, ScanReport):
+            is_author = sub_scan_report.author.id == request.user.id
+            return is_author
+        else:
+            return False
+    except Exception:
+        return False
+
+
 def has_viewership(obj: Any, request: Request) -> bool:
     """Check the viewership permission on an object.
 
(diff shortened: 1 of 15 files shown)


app/api/mapping/permissions.py
```diff
@@ -39,6 +39,34 @@ def is_az_function_user(user: User) -> bool:
...
+        # it belongs to and check the user has permission to view it.
+        if sub_scan_report_query := SCAN_REPORT_QUERIES.get(type(obj)):
+            sub_scan_report = sub_scan_report_query(obj)
+        else:
+            sub_scan_report = obj
+
+        # Check if the user is the author
+        if isinstance(sub_scan_report, ScanReport):
+            is_author = sub_scan_report.author.id == request.user.id
+            return is_author
```
dcl10, line 63:
```suggestion
            return sub_scan_report.author.id == request.user.id
```

app/next-client-app/components/HandleArchive.tsx
```diff
@@ -21,7 +21,7 @@ export const HandleArchive = async ({
       response = await archiveDataSets(id, !hidden);
       break;
     case "scanreports":
-      response = await updateScanReport(id, "hidden", !hidden);
+      response = await updateScanReport(id, { hidden: !hidden });
```
dcl10, line 24: I don't like how this reads. If `hidden` is `true`, meaning that something should be hidden, this code implies that in fact it should NOT be hidden.

dcl10 (changes requested):

app/api/mapping/permissions.py
AndrewThien: Aren't they the same? I thought return is_author would be also shorter. I don't fully understand. sorry

dcl10: So you're assigning the results of the comparison to a variable `is_author` and then immediately returning that variable. You may as well just return the comparison. It will do the same thing.

AndrewThien: Ah, I got what you mean when looking at the VS code, not on Github. Pushed the change. Thanks

app/next-client-app/components/HandleArchive.tsx
AndrewThien: so, when users click Archive/unarchive, the function will do its job which is receiving the current value of "hidden" (false/true) and sending to the backend the opposite value (true/false). in terms of logic, this code is doing well.
You don't like the naming or something else?

dcl10: I see what you mean now. Basically it kind of felt a bit odd on reading. But if it's doing what you expect then fine

dcl10 (approved):

Dev-Akashili (approved):

PR #794, opened 2024-07-15
Using Formik for Update Scan Report Table view

### LINKED ISSUE #791: Rebuild Update Table form
The last solution for building the Update Table page wasn't effective. So it should be rebuilt using Formik

AndrewThien:
# Changes

In the past, this view was built by using normal form fields and many other "unnecessary" components. 
This PR optimised the view of the Updating Scan Report Table by using Formik form and React Select. This makes the code shorter, cleaner, and easier to read.

This PR also added and modified some tooltips and toast messages.

Closes #791 

# Screenshots

General view

Screenshot 2024-07-15 at 10 14 42

# Checks

- [x] Run migrations, if any.
- [x] Run unit tests.


app/next-client-app/components/scanreports/ScanReportTableUpdateForm.tsx
```diff
@@ -0,0 +1,124 @@
...
+  };
+
+  return (
+    <Formik
+      initialValues={{
+        dateEvent: initialDateEvent[0].value,
+        personId: initialPersonId[0].value,
+      }}
+      onSubmit={(data) => {
+        handleSubmit(data);
```
Jivanam, line 65: should this not be asynchronous?

AndrewThien: I think handleSubmit should be an async function, to enable await update SR table to be done, then show the according toast (which is from lines 38-56)

Jivanam: I mean since `handleSubmit` is asynchronous, should we not use `async` and `await` in `onSubmit`?

AndrewThien: I see.
I think because onSubmit doesn't contain any code that needs to wait for handleSubmit to be complete, and I am handling the result of HandleSubmit inside itself (don't need to return a response, then handle this on onSubmit), async and await are not needed on OnSubmit.

Jivanam (approved):

app/next-client-app/api/scanreports.ts
```diff
@@ -137,18 +137,34 @@ export async function getScanReportTable(id: string): Promise<ScanReportTable> {
...
-    return null;
-  }
-
+export async function getScanReportField(id: string): Promise<ScanReportField> {
   try {
     return await request<ScanReportField>(fetchKeys.fieldName(id));
   } catch (error) {
-    console.warn("Failed to fetch data.");
-    return null;
+    console.warn("Failed to fetch data. Passed ID could be null");
```
blargly, line 144: If you're failing to retrieve data, why are we not erroring or redirecting? Creating a large object with empty strings, nulls and 0's doesn't feel like the right approach here.

blargly: Also, specifically in this case why are we returning a big object with "default" values? Is this supposed to be an empty form with no details? If so, do we need to really set the page to not display data?

blargly: Just to follow up - chatted with Tien (and Jon) about this - it looks like this follows a pattern already established in the app, so its beyond the remit of this PR to refactor, however the point still stands:

The API functions that hit the back end and then return an object should be neutral - they either find the object and return it, or it does not (e.g. a null). It is then up to specific pages that use that method how to handle that response. It feels strange to go to the trouble of building an "empty" object that is populated by the default values. If a particular view needs it, it should do that itself. 
In this case, the relevant view did not need the object prepopulating, and could easily handle a null (it was for a drop down, and to specific which value was selected).

Worth putting on the backlog as a bit of technical debt.

blargly (approved):
Please create an item of discussion in the back end for the comments.

AndrewThien:
> Please create an item of discussion in the back end for the comments.

Thanks for the comments. Please check out this newly created issue in backlog: https://github.com/Health-Informatics-UoN/Carrot-Mapper/issues/795

PR #798, opened 2024-07-23
Summary Rules modal for Mapping rules page

### LINKED ISSUE #751: Scan Report Mapping Rules Modals
# Summary

Rewrite the Scan Reporting Mapping Rules modals in Nextjs, depends on #750 

# Tasks
- [ ] Add "Show Summary View" modal
     - [ ] Add "simple" table rendering the summary data
     - [ ] This data is filtered on `source_concept_id` and only includes Rules `destination_field.name` that contain it. 
- [ ] Add "Analyse Rules" modal 
     - [ ] Find an example of "Analyse Rules" being populated (in dev, test, or production)
     - [ ] Mirror the existing view (this is unknown until we see one) 
     - [ ] Add "simple" table representing the Rules Analysis

# Acceptance Criteria

- [ ] Working modals to represent the Scan Report Mapping Rules

AndrewThien:
# Changes

This PR added the Summary rules modal for the Mapping rules page. Some things that are noticeable about this might be:

- The filtering process happened in the backend, not frontend (like the old version), meaning all the rules will be summarised first, then using a separate server-side pagination to get the data and present it to the UI.

-  The dialog modal is rendered on and by the intercepting route, which enabled the modal to show on top of the page while facilitating the server-side pagination with a separate route. Also, thanks to this implementation, users can easily copy and share the direct link to a segment of the modal. This feature may not be used in the future due to the new layout, but it is still a good opportunity to learn something new and useful.   

- The modal for the Analyse Rules function is not implemented here due to its "dangerous" and "expensive" features. So, if this PR gets approved, it can close #751, but I suppose we need to create another separate issue for the Analyse Rule function in the future. @AndyRae 

# Screenshots

Screenshot 2024-07-23 at 10 19 52

# Checks

- [x] Run migrations, if any.
- [x] Run unit tests.


app/api/api/views.py
```diff
@@ -1228,6 +1228,57 @@ def list(self, request):
         return Response(data={"count": count, "results": rules})
 
 
+class SummaryRulesList(RulesList):
+    def list(self, request):
+        # Get p and page_size from query_params
+        p = self.request.query_params.get("p", None)
+        page_size = self.request.query_params.get("page_size", None)
```
dcl10, line 1235: what happens if these 2 variables are `None`?

app/next-client-app/api/mapping-rules.ts
```diff
@@ -5,14 +5,15 @@ const fetchKeys = {
...
-  filter: string | undefined,
+  filter: string | undefined
 ): Promise<PaginatedResponse<MappingRule>> {
   try {
     return await request<PaginatedResponse<MappingRule>>(
-      fetchKeys.mappingruleslist(filter),
+      fetchKeys.mappingruleslist(filter)
     );
   } catch (error) {
     console.warn("Failed to fetch mapping rules.");
```
dcl10, line 19: We don't need this

app/next-client-app/api/mapping-rules.ts
```diff
@@ -54,3 +55,16 @@ export async function getMapDiagram(
...
+
+export async function getSummaryRules(
+  filter: string
+): Promise<PaginatedResponse<MappingRule>> {
+  try {
+    return await request<PaginatedResponse<MappingRule>>(
+      fetchKeys.summaryRules(filter)
+    );
+  } catch (error) {
+    console.warn("Failed to fetch summary rules data.");
```
dcl10, line 67: Nor here

dcl10 (changes requested):

dcl10 (approved):

AndyRae (approved):

PR #823, opened 2024-08-07
NextJS Docker containerising

### LINKED ISSUE #804: Containerise the Next.js project
## Tasks  

- [ ] New Dockerfile
- [ ] Mount the current code directory for the `app/next-client-app`
- [ ] Install npm dependencies
- [ ] Add to `docker-compose`
- [ ] Move `.env` configuration upwards to the Docker Compose
    - [ ] Web backend config
    - [ ] Nextjs .env config  

Use existing templates

AndrewThien:
# Changes

This PR added a Dockerfile for the NextJS app of Carrot, then connected it with the other Docker containers using docker-compose. It also:
- Get the web image up and running locally again
- Add omop-lite image to create "omop" schema on the DB, then connect it with other containers
- Fix the logo size
- Add Sharp package for NextJS

Closes #804 

# Checks

- [x] Run migrations, if any.
- [x] Run unit tests.


Jivanam (approved):

app/entrypoint.sh
```diff
@@ -2,17 +2,19 @@
...
 
-# Collect static files for serving
 cd /api
+
+# Collect static files for serving
 rm -rf staticfiles
 mkdir staticfiles
-python /api/manage.py collectstatic
-python /api/manage.py migrate
+python manage.py collectstatic
```
AndyRae, line 16: I think this all needs testing - I am hesitant to change anything in `entrypoint.sh` given it is used by the deployment as well. 
Ideally the only changes would be in `Dockerfile`, not `entrypoint`

docker-compose.yml
```diff
@@ -17,17 +18,59 @@ services:
...
+      - NEXTJS_URL=http://next-client:3000
+      - ALLOWED_HOSTS=['localhost', '127.0.0.1','web']
+      - COCONNECT_DB_ENGINE=django.db.backends.postgresql
+      - COCONNECT_DB_HOST=db
+      - COCONNECT_DB_PORT=5432
+      - COCONNECT_DB_NAME=omop
+      - COCONNECT_DB_USER=postgres
+      - COCONNECT_DB_PASSWORD=postgres
+      - COCONNECT_DB_AUTH_TOKEN=
+      - UPLOAD_ONLY=false
```
AndyRae, line 61: a couple of these are no longer needed:

```
COCONNECT_DB_AUTH_TOKEN
UPLOAD_ONLY
NLP_QUEUE_NAME
SCAN_REPORT_QUEUE_NAME
CREATE_CONCEPTS_QUEUE_NAME
CCOM_APP_URL
```

I would check `api/config/settings.py` for some that might be missing - we have sensible defaults that mostly assume the app is running on localhost.

app/entrypoint.sh
AndrewThien: Yeah, I see your point.
I left the entrypoint untouched and tested everything again. Everything worked fine on my end. However, it would be really helpful if you could spare some time to test the docker-compose on your end as well. Thanks

docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-name: carrot-mapper-dev
+name: carrot-mapper-dev-v2
```
AndyRae, line 1: I don't think this needs to be versioned in this way

AndyRae (changes requested):

app/entrypoint.sh
AndyRae: Perfect - thanks for that, this should work with the existing deployment then. It runs from scratch pulling it down for me, great job,

AndyRae (approved):

Jivanam (approved):

PR #881, opened 2024-09-12
Add "value_as_number" or "value_as_string" field to OBSERVATION table

### LINKED ISSUE #647: Add OBSERVATION table fields

- [X] I have searched the existing issues

### Is your proposal related to a problem or functionality gap?

The SDE programme has defined a minimum OMOP data model for Cohort Discovery. Observation-related data is included as it is relevant to research in the domain. It would be great if the following fields were included in the capability of CaRROT-Mapper.

### Describe your proposal

Please could you add the following fields from the OBSERVATION table? 

value_as_number
value_as_concept_id

### Describe alternatives you've considered

_No response_

_No response_

### Anything else?

_No response_

### Are you willing to contribute to developing this feature?

🎨 Yes, I'm keen to be involved in design discussions.

AndrewThien:
# Changes

This PR added a logic to create a mapping rule to the field "value_as_number"/"value_as_string" of the table OBSERVATION, based on the Scan Report Field's Data type, i.e., 

- when the concept's domain is Observation and Field's data type is INT or REAL, a mapping rule to "value_as_number" will be added along with other rules 
- when the concept's domain is Observation and Field's data type is VARCHAR, a mapping rule to "value_as_string" will be added along with other rules 

This PR also updated the "uses: actions/upload-artifact" from v2 to v3 in github/workflows/ci.yml

Notice: Users are not expected to add an "Observation" concept to the field which is not "VARCHAR"/"REAL"/"INT". In that case, adding error will appear with the information

Closes #647

# Checks

- [x] Run migrations, if any.
- [x] Run unit tests.


app/api/api/views.py
```diff
@@ -515,6 +515,18 @@ def post(self, request, *args, **kwargs):
         content_type_str = body.pop("content_type", None)
         content_type = ContentType.objects.get(model=content_type_str)
         body["content_type"] = content_type.id
+        concept_id = body.get("concept", None)
```
AndyRae, line 518: Could this section be moved down, to below the validation of whether a Concept exists? (Line `558`)

As you're then accessing the domain from the concept on line `519`, but it might not exist, so best to validate it.

This also removes the double lookup by to the Concepts table.

brian-kim31 (approved):

brian-kim31 (approved):

AndyRae (approved):

PR #896, opened 2024-10-10
Splitting out the Status field of a ScanReport

### LINKED ISSUE #696: Scan Report Index - Split out Status Columns
# Summary

While the investigation into the Azure based routing is ongoing, we can proceed with splitting out the column functionality. Status now should be: Status and Estimated Mapping Progress. 

Status is split into Uploading, Upload Failed, Awaiting Table Ids, Processing, Failed Processing, Done. This is set automatically by the app, depending on the upload and processing progress.

Mapping Progress is: 0%., 25%, 50%. 75% and Complete. Defaults to 0. This is set manually by the user. This the column used to check for "Mapping Complete".

We will need 2 new entities, and to seed these in our back end. Do this in the proper way - using Django Fixtures (already in the app_).

# Tasks

- [ ] 2 new entities - NewStatus and MappingProgress.
- [ ] Update ScanReport entity to have 2 new values (the above new entities)
- [ ] Seed the values for these
- [ ] Migrations
- [ ] Serialiser - update this to pull in the list options for mapping.
- [ ] Front end - add the 2 new columns
- [ ] add new properties to ScanReport type
- [ ] Hide Status (old column)
- [ ] Update ScanReport end point to set NewStatus when: file uploads, when file upload fails, when upload finishes, when processing starts, when processing fails, when processing finishes.
- [ ] Review

# Acceptance Criteria

- [ ] Table view, now no longer has old status column, but 2 new columns.
- [ ] (new) status shows new options depending on scan report functions - can't be set by user
- [ ] Mapping is still a drop down with a more limited range than before.

AndrewThien:
# Changes
This PR splits out the Status field of a ScanReport into Mapping status and Upload Status, then splits the column Status of the SR index table accordingly. One more filter for the Mapping status was also added to the UI of the SR index. Users can also see and change the Mapping status on the SR layout.

To make the splitting process safe, the PR creates a migration function to transfer the current data in the Status field to the two new fields before deleting the old Status field. Then, the Django backend and Azure functions are modified to match the changes in the Status field of an SR.

Lastly, tooltips for SR upload status were added as well as doing some code cleaning and moving icons of buttons (to enhance the consistency in UI)

Closes #696 

# Screenshots

Screenshot 2024-10-16 at 13 35 03

Screenshot 2024-10-10 at 12 12 37


app/next-client-app/constants/scanReportStatus.ts
```diff
@@ -1,47 +1,61 @@
...
+    color: "text-red-500 dark:text-red-500",
+  },
+  {
+    label: "Upload In Progress",
+    icon: Loader2,
+    value: "IN_PROGRESS",
+    color: "text-orange-500 dark:text-orange-500",
+  },
+];
+// Values of id and value here have to matched 100% with the fixtures in uploadstatus and mappingstatus tables in DB
```
AndyRae, line 23: How come `id` has to match? Makes me nervous matching on database ids.

app/next-client-app/components/scanreports/MappingStatus.tsx
```diff
@@ -7,58 +7,63 @@ import {
...
-  const statusInfo = statusOptions.find((option) => option.value === status);
+  const statusInfo = MappingStatusOptions.find(
+    (option) => option.value === mapping_status.value
+  );
   const textColorClassName = statusInfo?.color ?? "text-black";
 
   const handleChangeStatus = async (newStatus: string) => {
     const response = await updateScanReport(parseInt(id), {
-      status: newStatus,
+      // tried updating data wiht "value" but not successful, needed the id of the new mapping status here
```
AndyRae, line 37: Is this why you need the Id? If so - we should change the API to allow for values in the mapping status, not need ID.

app/api/config/react.py
```diff
@@ -1,7 +1,2 @@
-from shared.mapping.models import Status
-
-
 def react(request):
-    return {
-        "status": [{"id": id, "label": label} for id, label in Status.choices],
-    }
+    return {}
```
AndyRae, line 2: This is a breaking change, but I guess the PR is.. as in this is the first piece of functionality that will actually break the legacy React app. 
We should explore how much effort it would take to enable it to continue to work and avoid a breaking change.

app/workers/shared_code/db.py
```diff
@@ -19,7 +21,24 @@
...
+class MappingStatusConstant(Enum):
+    PENDING = "Pending Mapping"
+    COMPLETE = "Mapping Complete"
+    BLOCKED = "Blocked"
+    MAPPING_25PERCENT = "Mapping 25%"
+    MAPPING_50PERCENT = "Mapping 50%"
+    MAPPING_75PERCENT = "Mapping 75%"
+
+
+def update_scan_report_status(
```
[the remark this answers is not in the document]
AndrewThien: So when we upload a new SR, without setting the mapping_status along with upkoad_status, it will be null, and cause an error in the front-end app. To avoid this then, I will set the null check in Mapping Status component and set it to "PENDING" if it is null. Thanks ☺️

app/api/config/react.py
AndrewThien: I see. Reverted with some relevant updated.

AndyRae (approved):

AndyRae (approved):

brian-kim31 (approved):

PR #899, opened 2024-10-16
Add Projects pages: Index and Details

### LINKED ISSUE #607: View datasets by project
Could it be possible to group datasets by project, it can be hard to find all the connected datasets. It would be great to have a "Alleviate" view for example. 

### Breakdown

This requires 2 new pages; a list of all projects, and a Project detail view - that shows the datasets attached to it.

For the backend, there is an existing list all projects API, but this would require adding Projects to the `DatasetAndDataPartnerListView` serialising, and filtering.

For the frontend - this needs more fetching code for the Projects list, but then 2 new `/projects` + `/projects/[id]` pages.

### LINKED ISSUE #892: Datasets filter by Project API

### Summary

Adding `Projects` to the dataset API, so that `datasets_data_partners` can be filtered by projects.

`?project_id__in=123`

### Acceptance Criteria

- [ ] A backend API for a list of datasets filtered to a project.
- [ ] Tests pass.

### Tasks

- [ ] Update `DatasetAndDataPartnerViewSerializer` serialiser code. In this instance, it is fine for just the ID of the project to be serialised.
- [ ] Update `DatasetAndDataPartnerListView` add `Projects` to the existing filtering.

### Confirm creation

- [x] This issue is ready

### LINKED ISSUE #893: Project List/Detail Pages

### Summary

We require 2 new pages; a list of all projects, and a Project detail view - that shows the datasets attached to it.

We can use the same pattern we currently have for Datasets, and Scan Reports. With a layout, and table structure.
Fetching code already exists for the endpoints to fetch projects, and to fetch datasets.

### Acceptance Criteria

- [ ] A page to show a list of the users projects
- [ ] A page to show a project detail, with its relevant datasets.

### Tasks

- [ ] New page for listing projects `/projects`
- [ ] New table for rendering the list of projects
- [ ] New layout for a project detail `projects/[id]/layout.tsx`
- [ ] New page for project detail
- [ ] Add `projects` to the app navigation

### Confirm creation

- [x] This issue is ready

AndrewThien:
# Changes

This PR added two more new views for Projects: A list of available Projects (which users are the members of) and the Project detail page, which shows the datasets attached to the selected project, as well as other information.

Closes #892 
Closes #893 
Closes #607 

# Screenshots

Screenshot 2024-10-16 at 13 49 43
Screenshot 2024-10-16 at 13 49 49


app/next-client-app/api/projects.ts
```diff
@@ -0,0 +1,58 @@
...
+  } catch (error) {
+    console.warn("Failed to fetch data.");
+    return { count: 0, next: null, previous: null, results: [] };
+  }
+}
+
+export async function getAllProjects(): Promise<Project[]> {
+  try {
+    // If there are more than 100 projects in Carrot, adjust the page size below
+    return await fetchAllPages<Project>(fetchKeys.list("page_size=100"));
```
AndyRae, line 26: `page_size` should be an optional argument to this function instead of hardcoding it here.

app/next-client-app/api/projects.ts
```diff
@@ -0,0 +1,58 @@
...
+
+const fetchKeys = {
+  list: (filter?: string) => (filter ? `projects/?${filter}` : "projects/"),
+  projectsDataset: (dataset: string) => `projects/?dataset=${dataset}`,
+  project: (id: string) => `projects/${id}/`,
+};
+
+export async function getProjectsList(
+  filter?: string | undefined
+): Promise<PaginatedResponse<Project>> {
```
AndyRae, line 14: All these projects endpoints return the same type of `Project`, is this actually true? 
Looking at the api code it doesn't look like it.

For example `getProject` returns `ProjectSerializer` but the list view has 3 different serializer possibilities.
It's important we match up the return types to the API code accurately, so when those models are implemented on pages we can be safe that they actually contain that information.

AndyRae (changes requested):

app/next-client-app/api/projects.ts
AndrewThien: For some reason, the call will fail if the filter is empty. So for now, I will put an empty param here. Need more guidance here, actually

app/next-client-app/api/projects.ts
AndrewThien: so only `ProjectDatasetSerializer`  will return (therefore require) a different type with other serialiser/API calls (I mean, `getProjectsList` `getProjectsDataset` `getProject` will both hit `ProjectWithMembersSerializer` ), but it hasn't been implemented yet, I suppose.
If what I am thinking is right, so maybe I only need to update the type here?

app/next-client-app/api/projects.ts
AndrewThien: I found out that the function `fetchAllPages` was designed to expect the passed URL to already have a param query character there (`?`). So if the param for `fetchAllPages<Project>(fetchKeys.list());` is empty, `fetchAllPages` will call the URL `api/projects/&p=1` (without `?`) which will fail.

So I think we can either modify the function `fetchAllPages` (and we then need to update other related functions/components using this) or put an empty param for it (like " " or "p") as a "fake" filter (then it will call the URL with `?`)

AndyRae (approved):

PR #918, opened 2024-11-29
Add workers functions to create/update jobs' stage and status

AndrewThien:
# Changes

This PR added `create` and `update` workers functions for jobs' records along with some new types in workers code.
  
These functions were then applied to the existing Azure functions' workflow to create or update the record of the job's stage or stage's status in the database.

Notice: This PR aims to add basic workers functions to create and update jobs' stage and status, in order to get the basic details also information about the job's creation time and updated time. In the future, if more details about the job's progress/error handling are needed, these can be added later based on these basic functions.


app/workers/RulesOrchestrator/__init__.py
```diff
@@ -25,32 +31,53 @@ def orchestrator_function(context: df.DurableOrchestrationContext):
...
+
+        # Get concepts number
+        concepts_count = find_existing_concepts_count(table_id)
+        logger.info(f"Concepts found: {concepts_count}")
+
+        # Paginate, but ensure we have at least 1 task.
+        page_size = int(os.environ.get("PAGE_SIZE", "1000"))
+        num_pages = max((concepts_count + page_size - 1) // page_size, 1)
+
+        create_or_update_job(
```
AndyRae, line 49: Yep so the API should always be responsible for creating the job, workers only update.

AndyRae: (unless I've not thought of something that you have!)

AndrewThien: Yep, I just wondered if workers is only responsible for Job updating, how can it do that for the old SR tables which didn't have the Job record before? For example, if users want to re-run the functions of a current table by updating the table Person ID/Date event, the workers will try to update a job that didn't exist and there will be errors, I think.

AndrewThien: So should I make the backend create initial job records (when uploading a new SR, initial records for 5 stages will be created for SR and Sr tables), and also in Workers, a function to create_or_update jobs, for the sake of re-running the functions (of both old and new SR tables)

AndyRae: > For example, if users want to re-run the functions of a current table by updating the table Person ID/Date event, the workers will try to update a job that didn't exist and there will be errors, I think.

We would create a new "job" for every run of the updating the table. The idea being a job is created for every unique run of the workers.

> So, when users upload a new SR, I think we can make the API to create 2 initial Job records for the upload and download stages. For each SR table, 3 initial Job records for each stage will be created as well at this time. But this needs to be done in workers anyway, cause the logic to create SR tables is currently in there.

So this wouldn't be necessary to create initial records and update them - we need to create the jobs when the backend is triggering the workers every time. It isn't the case that every SR has a limited number of jobs, it is unlimited as it's for every single run of the workers.

AndyRae (approved):

Dev-Akashili (approved):

Jivanam (approved):

PR #925, opened 2024-12-18
JobsData type bug fix

AndrewThien:
# Changes

This PR provided a small fix for jobsData by adding type safeguarding before passing it to the FindGeneralStatus function.

Related #923

### DIFF
--- a/app/next-client-app/components/jobs/JobUtils.tsx
+++ b/app/next-client-app/components/jobs/JobUtils.tsx
@@ -1,21 +1,19 @@
 export const FindGeneralStatus = (jobsData: Job[]) => {
-  let generalStatus = "NOT_STARTED";
-  if (jobsData.length > 0) {
-    if (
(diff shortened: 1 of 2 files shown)


app/next-client-app/components/jobs/JobUtils.tsx
```diff
@@ -1,21 +1,19 @@
...
-      generalStatus = "FAILED";
-    } else if (
-      jobsData?.every((job) => job.status && job.status.value === "COMPLETE")
-    ) {
-      generalStatus = "COMPLETE";
-    }
+  if (
+    jobsData.some((job) => job.status && job.status.value === "IN_PROGRESS")
+  ) {
+    return "IN_PROGRESS";
```
AndyRae, line 5: Happy to approve to get this fix in - but looks like we have configuration masking as constants, and not benefiting from any typing. 

As in - lots of strings being passed about when we'd really like types.

AndyRae (approved):

PR #930, opened 2025-01-09
Fix downloading files, SR exporting and other bugs

### LINKED ISSUE #926: Downloading files is broken

### Current Behavior

Download a Scan Report, or rules does not work.

### Expected Behavior

_No response_

### Steps To Reproduce

Upload a scan report
Try to download it

### Environment

```markdown
- OS:
- Other environment details:
```

_No response_

### Anything else?

The code to download files would previously link directly to the API that would allow for the file to be downloaded - the endpoint was hardcoded in the html.
This no longer works as the url has changed, but additionally won't work as the user client won't be authenticated to download it.
Effectively the Next js server is authenticated, and needs to download the file and return to the browser

AndrewThien:
# Changes

This PR fixes the process of downloading files and ScanReport exporting in Carrot as well as other bugs that may have been caused by separating the Front end from the Back End. 

This PR also add the Export Poetry Plugin to the DockerFiles and workflows.

Closes #926

### DIFF
--- a/app/next-client-app/app/(protected)/scanreports/[id]/downloads/columns.tsx
+++ b/app/next-client-app/app/(protected)/scanreports/[id]/downloads/columns.tsx
@@ -12,8 +12,10 @@ import {
   TooltipProvider,
   TooltipTrigger,
 } from "@/components/ui/tooltip";
-import Link from "next/link";
+import { saveAs } from "file-saver";
 import { Badge } from "@/components/ui/badge";
+import { downloadFile } from "@/api/files";
+import { toast } from "sonner";
 
 export const columns: ColumnDef<FileDownload>[] = [
   {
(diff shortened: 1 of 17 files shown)


MoizIqbal97 (dismissed):

app/next-client-app/app/(protected)/scanreports/[id]/downloads/columns.tsx
```diff
@@ -82,15 +84,32 @@ export const columns: ColumnDef<FileDownload>[] = [
     id: "Download",
     header: ({ column }) => <DataTableColumnHeader column={column} title="" />,
     cell: ({ row }) => {
-      const { id, scan_report } = row.original;
+      const { id, scan_report, file_type, name } = row.original;
+      const handleDownload = async () => {
+        const response = await downloadFile(scan_report, id, file_type.value);
```
Comment (Jivanam) [inline app/next-client-app/app/(protected)/scanreports/[id]/downloads/columns.tsx:89]: Can we call `downloadFiles` which I believe is server method from the client component?

Jivanam (dismissed):

app/next-client-app/app/(protected)/scanreports/[id]/downloads/columns.tsx
```diff
@@ -82,15 +84,32 @@ export const columns: ColumnDef<FileDownload>[] = [
     id: "Download",
     header: ({ column }) => <DataTableColumnHeader column={column} title="" />,
     cell: ({ row }) => {
-      const { id, scan_report } = row.original;
+      const { id, scan_report, file_type, name } = row.original;
+      const handleDownload = async () => {
```
Comment (AndyRae) [inline app/next-client-app/app/(protected)/scanreports/[id]/downloads/columns.tsx:88]: This fixes the `downloads` page, but not the `Export Scan Report` though right?

Comment (AndrewThien) [reply app/next-client-app/app/(protected)/scanreports/[id]/downloads/columns.tsx:88]: yes, I forgot about that. Let me fix that in this PR as well. Thanks!

Jivanam (approved):

Issue #979, opened 2025-03-12
Long processing time for table edit feature


erummas:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

The mapper is taking longer than usual to process after the ID and date is set for each table. It used to be a fairly quick process. This is not an urgent request but something to look into. 

### Environment

```markdown
- OS:
- Other environment details:
```

### Are you willing to contribute to resolve this issue?

None

AndrewThien:
Hi Erum,
Thanks for creating the issue.
Our team is investigating the problem and tracking the functions' performance. Hopefully will have some updates on this soon.
Cheers!

AndrewThien:
Hi @erummas , in the latest release, we did some logic optimisation and had some better processing speed related to rules generation. Can you please try again to see if there is any improvements related to this issue on your end?
Thanks!

PR #986, opened 2025-03-18
Optimise Rules creation logic

AndrewThien:
| |
|-|
♻️ Refactor
⚡️ Optimization

## PR Description
This PR optimised the logic and speed of Carrot's worker function RulesGenerationActivity / code in `shared/services/rules.py`. Refer to the table below for the speed difference taken by RulesGenerationActivity of the table having 12723 concepts added (unit: minutes):

| Mode  | Before PR | After PR |
|-------|-----------|----------|
| Local | 17        | 12       |
| Dev   |      |    |
| Test  |     |      |

Related #979


app/shared/shared/services/rules.py
```diff
@@ -448,33 +452,33 @@ def _save_mapping_rules(scan_report_concept: ScanReportConcept) -> bool:
         )
         rules.append(rule_domain_value_as_string)
 
-    # now we are sure all rules have been created, we can save them safely
-    for rule in rules:
-        rule.save()
-
-    return True
+    return True, rules
```
AndyRae, line 455: Looks like we could simplify and just return the rules?

app/api/api/views.py
```diff
@@ -639,8 +639,8 @@ def post(self, request, *args, **kwargs):
         self.perform_create(serializer)
 
         model = serializer.instance
-        saved = _save_mapping_rules(model)
-        if not saved:
+        rules = save_mapping_rules(model)
+        if rules == []:
```
AndyRae, line 643: `if not rules`
would work and be cleaner - an empty list in python is falsy

AndrewThien: thought it only works for boolean. :D will change it now

AndyRae (approved):

brian-kim31 (approved):

PR #987, opened 2025-03-18
Storage code refactor 

### LINKED ISSUE #970: Preparing: Set up Docker compose and Refactoring

### Summary

Preparing steps for adding Minio as a storage option for Carrot Mapper

### Acceptance Criteria

- [x] Carrot has central functions to interact with Storage
- [x] Interactions with Azure storage are still stable on local dev. and dev.carrot.ac.uk and test.carrot.ac.uk

### Tasks

- [x] Initial refactor code: Make the `central` services where blob storage connection and related functions are set and used: from `app/workers/shared_code/blob_parser.py` and `app/shared/shared/files/service.py`
- [x] Put the central services here: shared/services/storage_service.py

### Confirm creation

- [x] This issue is ready

Karthi-DStech:
### PR Description

This PR introduces Centralised Logic for the storage and Automatic Queue and Blob Container. Below are the key changes and instructions for testing:

- **Centralized Storage Logic:** Centralised Storage logic from `blob_parser.p`y and `service.py` to `storage_router.py`. 

- **Code Refactoring:** Refactored and removed redundant code. 

### Instructions for checking the code

1. Clone the branch and locate it in the root directory. 

2. Clear Persistent Storage (Optional):

- If you need to clear the persistent storage, follow the code or skip the process

```
docker compose down --volumes --remove-orphans

docker system prune -a --volumes

```

3. Start the Application:

```
docker compose up -d
```

4. Create a Superuser and Create Dataset, Data Partner and Projects. 

5. Add the listed blob containers and queue to the Microsoft Storage Explorer.

6. Add the mappings and test it. 

For more information, follow the link: https://carrot.ac.uk/mapper/dev_guide/quickstart

#### Test Cases

- Conducted comprehensive testing of all functions, ensuring successful execution without errors throughout the upload and download processes.

- [✅] I've completed all actions and tasks detailed in the PR Template

closes #970


app/shared/shared/services/storage_router.py
```diff
@@ -0,0 +1,287 @@
...
+    def upload_blob(
+        self,
+        blob_name: str,
+        container: str,
+        file: Union[bytes, str, Iterable[AnyStr], IO[AnyStr]],
+        content_type: str,
+        use_read_method: bool = False,
+    ):
+        """
+        This function takes a file and uploads it to a specified container in Azure Blob Storage.
```
AndrewThien, line 212:
```suggestion
        This function takes a file and uploads it to a specified container in Data Storage.
```

app/shared/shared/services/storage_router.py
```diff
@@ -0,0 +1,287 @@
...
+        The file is stored with the provided blob name and content type.
+
+        Args:
+            blob_name (str): The name that will be assigned to the uploaded file in Azure Blob Storage.
```
AndrewThien, line 216:
```suggestion
            blob_name (str): The name that will be assigned to the uploaded file in Data Storage.
```

app/shared/shared/services/storage_router.py
```diff
@@ -0,0 +1,287 @@
...
+            container (str): The name of the Azure Blob Storage container where the file will be uploaded.
```
AndrewThien, line 217:
```suggestion
            container (str): The name of the Data Storage container where the file will be uploaded.
```

app/shared/shared/services/storage_router.py
```diff
@@ -0,0 +1,287 @@
...
+    def _read_file(self, file, use_read_method: bool):
+        if use_read_method:
+            read_file = file.read()
+            return read_file
+
+        else:
+            open_file = file.open()
+            return open_file
+
+    def upload_form_from_data_dictionary_to_containers(
```
AndrewThien, line 247: do we need to create a separate function to do this? this is only used once so maybe it is not that necessary.

app/shared/shared/services/storage_router.py
AndrewThien, line 1: can the file name be `storage_service`? `router` would be associated with linking and moving between things in my mind :D 
Need more opinion here @AndyRae

app/shared/shared/mapping/views.py
```diff
@@ -41,10 +39,13 @@
     get_mapping_rules_json,
     make_dag,
 )
+from shared.services.storage_router import StorageService
 
 from .forms import ScanReportAssertionForm, ScanReportForm
 from .permissions import has_editorship, has_viewership, is_admin
 
+storage_parser = StorageService()
```
AndrewThien, line 47: can it be `storage_service`? `parser` would be associated with transforming/passing things in my mind :D
Need more opinion here I think @AndyRae

AndrewThien (changes requested):

app/api/test/conftest.py
```diff
@@ -54,7 +54,7 @@ def django_db_setup(
     # Manually create the test DB + omop Schema if it doesn't exist, drop if it does.
     run_sql("postgres", f"DROP DATABASE IF EXISTS {db_name}")
     run_sql("postgres", f"CREATE DATABASE {db_name} TEMPLATE template0")
-    run_sql(db_name, "CREATE SCHEMA IF NOT EXISTS omop")
+    run_sql(db_name, "CREATE SCHEMA omop;")
```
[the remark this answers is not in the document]
Karthi-DStech: Ah. I got my test cases to fail locally at some point. So I did make these changes. Let me have a look at it and try to see if I can remove it.

app/api/entrypoint.sh
```diff
@@ -11,5 +11,10 @@ python manage.py collectstatic
 python manage.py migrate
 python manage.py automatic_seeding_data
 
+# Check if STORAGE_TYPE is set to "azure", then setup storage
+if [ "$STORAGE_TYPE" = "azure" ]; then
```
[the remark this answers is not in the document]
Karthi-DStech: No problem. I just added this in the PR for discussing the idea. I can remove it in this PR and have a discussion with the team so see if we can make this an Issue. 

Thanks for the message Thien.

app/shared/shared/services/storage_router.py
Karthi-DStech: Initially there were two upload functions and I centralised them.

- This new function handles uploads by accepting a use_read_method parameter, eliminating the need for separate functions or internal conditional logic. 

- This approach simplifies the code and makes it more maintainable. By passing the appropriate parameter when calling `upload_blob`, we can control the file reading behaviour directly at the call site. 

However, if you have other ideas, please do let me know. I can make changes accordingly.

app/shared/shared/services/storage_router.py
Karthi-DStech: Hi Thien. Thanks for the message. 

Actually, I decided to use `storage_router.py` because, in the next PR, we will include MinIO which works with Azure. This centralised Logic will route the required storage and apply the logic as well. 

Any thoughts on this? 
@AndyRae @AndrewThien

app/shared/shared/services/storage_router.py
Karthi-DStech: Ok

app/shared/shared/services/storage_router.py
Karthi-DStech: Ok

app/shared/shared/services/storage_router.py
Karthi-DStech: Ok

app/api/api/serializers.py
```diff
@@ -12,18 +13,17 @@
 from shared.data.models import Concept
 from shared.mapping.models import (
     Dataset,
+    MappingStatus,
```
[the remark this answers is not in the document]
Karthi-DStech: Sure Andy, thanks for your message.

app/api/entrypoint.sh
```diff
@@ -10,7 +10,7 @@ mkdir staticfiles
 python manage.py collectstatic
 python manage.py migrate
 python manage.py automatic_seeding_data
-python manage.py default_super_user
```
AndrewThien, line 13: I don't think this removal is correct

app/shared/shared/services/storage_router.py
AndrewThien: I saw you changed the name but still want to clarify: Yes, in this central service, we can switch (or do the routing) between Azure and Minio. However, the routing is just an internal feature in this file and it mostly includes services related to storage. So I think the file name should reflect the main content of the file, which will prevent confusion for other (future) developers.

app/shared/shared/services/storage_router.py
AndrewThien: I am mentioning the function `upload_form_from_data_dictionary_to_containers`.  I think you are trying to replace lines 280-288 of the file `app/shared/shared/mapping/views.py` by creating and using this function. I personally don't think it is necessary because you can just reuse `upload_blob`, for example, to do this, instead of creating a new function in `storage_service`.

app/api/config/settings.py
```diff
@@ -239,8 +239,15 @@
...
 
 
-##
-# https://github.com/Health-Informatics-UoN/carrot-mapper/issues/946
-SUPERUSER_DEFAULT_PASSWORD = os.getenv("SUPERUSER_DEFAULT_PASSWORD", None)
-SUPERUSER_DEFAULT_EMAIL = os.getenv("SUPERUSER_DEFAULT_EMAIL", "user@carrot")
-SUPERUSER_DEFAULT_USERNAME = os.getenv("SUPERUSER_DEFAULT_USERNAME", None)
+# Storage Type Variable
+STORAGE_TYPE = os.getenv("STORAGE_TYPE", "azure")
+if not STORAGE_TYPE:
```
AndrewThien, line 244: I don't think this check is necessary cause STORAGE_TYPE is having a default value

AndrewThien:
Just a note for myself to add `STORAGE_TYPE` on env. vars across Azure deployments when this PR is ready to be merged.

app/shared/shared/mapping/views.py
AndrewThien, line 1: I think the view `ScanReportFormView` is overlapped with `ScanReportIndexV2` - Perform_create, so at least we don't need to do anything here. 
Should we delete this? @AndyRae

AndrewThien:
These are some questions for @AndyRae related to the use of env. vars in Django:

- Before this PR, in the `shared` app, only one env. var was needed (`STORAGE_CONN_STRING` for `add_message` function). I wonder where did the `shared` app get this variable from. I assumed it was from the `api/config/settings.py` because `api` was using it to send queue message, and on the Dockerfile, `shared` will be copied over to `api`, but `STORAGE_CONN_STRING` wasn't there.

- In this PR, `storage_service` in the `shared` app will need one more env var (`STORAGE_TYPE`) and will be used by `workers` and `api`, where should the app get this variable from? How can we make sure the variable will be retrieved correctly? I am asking because if I am not using Docker compose to test this PR, the `storage_service` will fail at line 22 in `shared/services/storage_service.py` because Django couldn't find `STORAGE_TYPE`, 
and if I do ```self._storage_type = os.getenv("STORAGE_TYPE", "azure")```, Django will be happy

app/shared/shared/mapping/views.py
AndyRae: Yeah I'm pretty sure this is code we missed in the deletion! Same with the `urls`

app/shared/shared/services/storage_router.py
AndyRae: `router` is very much a loaded word in Django - it would indicate a django rest framework api.

AndyRae:
> These are some questions for @AndyRae related to the use of env. vars in Django:
> 
> * Before this PR, in the `shared` app, only one env. var was needed (`STORAGE_CONN_STRING` for `add_message` function). I wonder where did the `shared` app get this variable from. I assumed it was from the `api/config/settings.py` because `api` was using it to send queue message, and on the Dockerfile, `shared` will be copied over to `api`, but `STORAGE_CONN_STRING` wasn't there.
It looks like it would read it straight from the environment? (which was not good!)

> * In this PR, `storage_service` in the `shared` app will need one more env var (`STORAGE_TYPE`) and will be used by `workers` and `api`, where should the app get this variable from? How can we make sure the variable will be retrieved correctly? I am asking because if I am not using Docker compose to test this PR, the `storage_service` will fail at line 22 in `shared/services/storage_service.py` because Django couldn't find `STORAGE_TYPE`,
>   and if I set the default for it here, which is not quite right, e.g., `self._storage_type = os.getenv("STORAGE_TYPE", "azure")`, Django will be happy

Good question. We could just read from the environment and be happy too - but you're right in that this feels bad. We're kind of in a bind here - as `workers` and `api` both define their own django settings, so anything in `shared` needs to import from them. Therefore we need a shared settings file... not great.
The ideal (longer term) resolution is - we drop `shared`, and `workers` uses the `api` as a library to connect to the DB. I think this should happen anyway, as it simplifies the codebase. 
Short term - I wouldn't mind it being read from the env as in your fix.

app/shared/shared/services/storage_router.py
Karthi-DStech: Got it. Thanks for your message @AndrewThien @AndyRae

app/api/test/test_setup.py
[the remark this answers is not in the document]
Karthi-DStech: There were some conflicts that prompted me to resolve the issue by merging it into the main branch. However, I later realized the changes weren’t necessary and could lead to further conflicts. 

So, I revoked the merge and manually fixed the issue instead.

app/api/config/settings.py
Karthi-DStech: Anyhow, I'll include this in the next PR. Should I want to remove it now and add it later or just let it be as it is Thien?

app/shared/shared/mapping/views.py
Karthi-DStech: Is there anything I could modify?

AndrewThien: let me handle this deletion @Karthi-DStech. I will push the related commit soon

app/shared/shared/services/storage_router.py
AndrewThien: since the `app/shared/shared/mapping/views.py` will be deleted, I don't think we need `upload_form_from_data_dictionary_to_containers`, right?

app/shared/shared/mapping/views.py
Karthi-DStech: Alright Thien.

app/api/test/test_setup.py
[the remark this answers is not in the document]
AndrewThien: so, do you want to put the merged commits from the last PR from Peter back?

app/api/config/settings.py
AndrewThien: I mean, if it has the default value, the condition `if not STORAGE_TYPE` will never be met. That's why I don't think it is necessary

app/shared/shared/services/storage_router.py
Karthi-DStech: In the next PR, I plan to remove the upload_form_from_data_dictionary_to_containers function and integrate its logic with the MinIO implementation. This is because most of the logic should be centralized to ensure compatibility with both MinIO and Azure. By addressing this in the next PR, I can find a more efficient and streamlined solution.

If there’s still any confusion about the requirements, we can schedule a quick catch-up at your convenience, @AndrewThien

app/shared/shared/services/storage_service.py
```diff
@@ -0,0 +1,287 @@
...
+    remove_BOM,
+)
+
+logger = logging.getLogger("test_logger")
+
+
+class StorageService:
+    def __init__(self):
+        self._client = None
+        self._storage_type = os.getenv("STORAGE_TYPE")
```
AndrewThien, line 22: is that working on your end without the default "azure" value? mine is not

app/api/entrypoint.sh
Karthi-DStech: Put back all the deleted files from peter's PR.

Karthi-DStech:
Fixed all bugs (Restored Superuser Creation Commit & Removed Error Message of STORAGE_TYPE). 

- Tested the code and works well without any errors. 

Let me know your thoughts and see if I can do anything. @AndrewThien @AndyRae

AndrewThien (dismissed):

AndrewThien (approved):

brian-kim31 (approved):

AndyRae (approved):

PR #1000, opened 2025-03-26
Feature Added - MinIO

### LINKED ISSUE #968: Add/modify configs

Karthi-DStech:
♻️ Refactor
✨ Feature

#968 **Feature Added - MinIO**

## PR Description
This PR implements a configurable storage backend system supporting both 
Azure Blob Storage and MinIO. 

- [✅] Tested locally and worked without any errors. 

## Key Updates
- Implemented MinIO as Storage.
- Refactored Storage Service Logic.
- Renamed methods of the Storage Logic.
- Added MinIO to the poetry.

## Screenshots: Example Behaviour of MinIO

**Successfully Generated Mapping Rules and Downloaded**

Screenshot 2025-03-26 at 3 31 39 pm

**MinIO Interface for uploaded files in the container**

Screenshot 2025-03-26 at 3 32 50 pm

#### - Remove existing Docker cache and persistent volumes to ensure a clean setup (if necessary):

```
docker compose down --volumes --remove-orphans
docker system prune -a --volumes
```
#### - Rebuild and start the containers from scratch:
```
docker compose up --build
```

#### - If you are going to use MinIO, follow the commands below: 

1. Open Microsoft Storage Explorer and create the following Queue

```
rules-local
rules-exports-local
uploadreports-local
```
2. Access the MinIO Console at `http://localhost:9001` using the default credentials:
```
**Username:** minioadmin
**Password:** minioadmin
```

3. Create the required storage buckets:
```
scan-reports
data-dictionaries
rules-exports
```

4. Make sure you set-up the `STORAGE_TYPE=minio` in the env variables of the docker-compose for `api`and `workers` containers. 

5. Create Data Partner, Datasets and Projects and upload the file. 

#### - If you are going to test Azure Blob as a storage: 

1. Make sure you set-up the `STORAGE_TYPE=azure` in the env variables of the docker-compose for `api` and `workers` containers. 

2. Set up the Blob Queue and Blob Containers in the Microsoft Storage Explorer. 

## Feedback
Feel free to share any feedback or suggestions for further improvements. Your input is always welcome! 🚀

Closes #968


app/shared/shared/services/storage_service.py
```diff
@@ -200,46 +339,158 @@ def modify_filename(self, filename: str, dt: str, rand: str) -> str:
...
+        """
         if use_read_method:
             read_file = file.read()
             return read_file
 
         else:
             open_file = file.open()
             return open_file
+
+    def _upload_minio_to_rules_export(
```
AndrewThien, line 432: I don't think we need a separate function for this. We can combine it to `upload_file`, just do something similar to `use_read_method` or not, in case of `azure`, I suppose.

app/shared/shared/services/storage_service.py
```diff
@@ -200,46 +339,158 @@ def modify_filename(self, filename: str, dt: str, rand: str) -> str:
...
+                file_name,
+                data=file,
+                length=file_size,
+                content_type=content_type,
+            )
+        except Exception as e:
+            logger.error(f"Error uploading the file to MinIO: {e}")
+            raise ValueError(f"Error uploading the file to MinIO: {e}")
+
+    def upload_to_rules_export(self, file_name, file, content_type):
```
AndrewThien, line 465: If we can combine `_upload_minio_to_rules_export` to `upload_file` then we don't need this function as well (just add one more param, I think)

AndrewThien (changes requested):

app/shared/shared/services/storage_service.py
```diff
@@ -18,11 +19,21 @@
...
     def __init__(self):
+        """
+        Service for interacting with cloud storage
+        providers (Azure Blob Storage and MinIO).
+        """
         self._client = None
         self._storage_type = os.getenv("STORAGE_TYPE")
 
-    def _get_service_client(self):
+    def _get_service_client(self) -> Optional[Union[BlobServiceClient, Minio]]:
```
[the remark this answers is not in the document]
Karthi-DStech: Just to clarify, you're suggesting that we check the storage type and return an enum. Based on the enum value of the storage type, the corresponding storage logic should be implemented. Is that correct?

For example: 

# Make this to return an ENUM
def _get_client_type(self) -> StorageType:  
    return self._storage_type

 # Returns initialized client
def _get_client(self) -> Union[BlobServiceClient, Minio]: 
    if self._client is None:
        self._client = self._init_client()
    return self._client

# Eg of upload Blob
def upload_blob(self):
    client_type = self._get_client_type()
    client = self._get_client()
    
    if client_type == StorageType.AZURE:
        # Azure-specific logic
    else:
    # MinIO logic

docker-compose.yml
```diff
@@ -107,16 +110,19 @@ services:
...
       - DB_PASSWORD=postgres
-      - STORAGE_TYPE=azure
-      - PYTHONPATH=/home/site/wwwroot:/home/site/wwwroot/shared
+      - STORAGE_TYPE=minio
+      - MINIO_ENDPOINT=minio:9000
+      - MINIO_ACCESS_KEY=minioadmin
+      - MINIO_SECRET_KEY=minioadmin
     volumes:
       - ./app/workers:/home/site/wwwroot
-      - ./app/shared:/home/site/wwwroot/shared
```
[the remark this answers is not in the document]
Karthi-DStech: Ok. I'll revoke the changes.

app/shared/shared/services/storage_service.py
Karthi-DStech: Let me explain why we currently have separate upload methods.

The `upload_file` method uses `self._read_file` to either open the file using `.open()` or read the file using `.read()`.

Initially, I used `upload_file` to upload files to the `rules-exports` container using MinIO. However, I encountered an error indicating that MinIO could not perform `read` or **`open` operations on **BytesIO**-like file structures. 

After researching, I understood the differences between Azure Blob Storage and MinIO:

**Key Differences in Handling Files:**

- Azure Blob Storage: It can handle both `BytesIO` and raw bytes. The use of `.read()` is appropriate when uploading to the `rules-exports` container because it allows direct memory access to pre-loaded `BytesIO` buffers, avoiding file I/O.

- MinIO: It directly processes `BytesIO` buffers from memory and does not support `read` or `open` operations. MinIO’s SDK protocol requires the file length to ensure complete uploads, which cannot be directly calculated from `BytesIO`.

**Instead, it requires the following:**

- Manually specifying the file length
- An explicit `seek(0)` call before uploading
- It doesn't work with our current `_read_file` helper for `BytesIO` data.
```
file.seek(0)
file_size = file.getbuffer().nbytes
```

Given this, when using `upload_file` for `rules-exports`, it should not utilize `_read_file`.


**Why Not Add Another Parameter to Control This?**

- Adding more parameters would increase the complexity of the `upload_file` method.
- For now, keeping them separate results in cleaner, more maintainable code.
- There is potential to optimize the upload functions, and I plan to address this in a separate PR to minimize changes to the existing source code.


**Example of What Doesn't Work:**

If we tried using the current `upload_file` for MinIO rules exports, we'd either:
- Encounter length calculation errors, or
- Need to add complex conditional logic that would make the method harder to understand

Would you like to pair up to explore potential refactoring approaches? I'm happy to walk through the specific error cases I encountered during testing.

app/shared/shared/services/storage_service.py
Karthi-DStech: Ah I see. I get it now. 

If we combine `_upload_minio_to_rules_export` to `upload_file`, it should be fine. I can make a condition based on uploading a file to the rules-exports. 

Should I want to implement the changes? @AndrewThien

Karthi-DStech:
Thanks for your comments and I have replied to them @AndrewThien. Let me know your thoughts at your convenience.

app/shared/shared/services/storage_service.py
```diff
@@ -200,42 +322,109 @@ def modify_filename(self, filename: str, dt: str, rand: str) -> str:
...
-    def upload_blob(
+    def upload_file(
         self,
-        blob_name: str,
+        file_name: str,
         container: str,
         file: Union[bytes, str, Iterable[AnyStr], IO[AnyStr]],
         content_type: str,
         use_read_method: bool = False,
+        minio_rules_export: bool = False,
```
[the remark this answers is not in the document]
Karthi-DStech: Got it.

Karthi-DStech:
**Root Cause Analysis for Failing in Github Test Cases:**

- The existing setup works completely fine when Carrot-Mapper is running using `docker-compose`.
- Django test couldn't access the variables for MinIO. 
- Even though the variables are there in `settings.py`, the test cases fail. 
- If I hardcode the variables in `storage_service.py`(as shown below), it passes the test. 
```
self._storage_type = os.getenv("STORAGE_TYPE", "minio")
```

**Conclusion**

- Tests are getting passed when variables are hard-coded.  
- Carrot-Mapper runs successfully (both Azure and MinIO) without any errors in `docker-compose`
- Tests fail when pushing the code for the PR. 

Assistance Required - @AndyRae @AndrewThien @brian-kim31

app/shared/shared/services/storage_service.py
```diff
@@ -3,188 +3,326 @@
...
+        except Exception as e:
+            logger.error(f"Error connecting to Azure Blob Storage: {e}")
+            raise ValueError("Failed to initialise Azure Blob Storage client.")
+
+    def _initialise_minio_client(self):
+        """
+        Initialises the MinIO client.
+        """
+        try:
+            # Hard-coded MinIO credentials for testing purposes
```
AndrewThien, line 58: please remove this comment

Karthi-DStech: Ok👍🏻

brian-kim31 (approved):

PR #1015, opened 2025-04-15
Added validator for dd & sr

### LINKED ISSUE #1003: Improve Validation for Data Dictionary

### Is your proposal related to a problem or functionality gap?

Currently, when uploading ScanReport and Data dictionary, there are some validations for DD, but they are not enough.  Specifically, the table names in `csv_file_name` column of DD have to exactly match the table name in the column `Table` in SR (including the extension `.csv`). Users can only put only the name of the table, without `.csv` and the uploading still goes through, which will cause problems later for other processes.

### Describe your proposal

Add more validations inside `validate_data_dictionary_file` function in the `ScanReportFilesSerializer` to deal with the problem above.

### Describe alternatives you've considered

_No response_

yes

### Anything else?

_No response_

### Are you willing to contribute to developing this feature?

Karthi-DStech:
⚡️ Optimization #1003 

## PR Description

This PR implements **validation to check the Data Dictionaries' `csv_file_name` entries exactly match `Scan Report table names` (including file extensions)**. The changes prevent processing errors by catching mismatches during upload

## Key Changes

- Added validation in ScanReportFilesSerializer to:
    - Extract table names from Scan Report's "Table" column (including .csv extensions)
    - Compare against Data Dictionary's `csv_file_name entries`
    - Reject uploads with mismatched names

- Enhanced error messages that clearly show:
    - Which table names are invalid
    - Expected table names from the Scan Report

## Feedback
Feel free to share any feedback or suggestions for further improvements. Your input is always welcome! 🚀

Closes #1003

### DIFF
--- a/app/api/api/serializers.py
+++ b/app/api/api/serializers.py
@@ -1,6 +1,7 @@
 import csv
 from collections import Counter
 from io import BytesIO, StringIO
+from api.logger import logger
 
 import openpyxl  # type: ignore
 from datasets.serializers import DatasetSerializer
@@ -28,23 +29,66 @@ from config.settings import DATA_UPLOAD_MAX_MEMORY_SIZE
 
 class ConceptSerializerV2(serializers.ModelSerializer):
     class Meta:
+        """
+        Serializer for the Concept model.
+
+        Serializes the concept_id, concept_name, and
+        concept_code fields.
+
+        Args:
+            model (Concept): The model to be serialized.
+
+            fields (list): The fields to be included in
+            the serialized output.
+        """
+
         model = Concept
         fields = ["concept_id", "concept_name", "concept_code"]
 
 
 class UserSerializer(serializers.ModelSerializer):
+    """
+    Serializer for the User model.
+
+    Serializes the id and username fields.
+
+    Args:
+        model (User): The model to be serialized.
+
+        fields (list): The fields to be included in
+        the serialized output.
+    """
+
     class Meta:
         model = User
         fields = ("id", "username")
 
 
 class UploadStatusSerializer(serializers.ModelSerializer):
+    """
+    Serializer for the UploadStatus model.
+
+    Serializes the value field.
+    """
+
     class Meta:
         model = UploadStatus
         fields = ["value"]
 
 
 class MappingStatusSerializer(serializers.ModelSerializer):
+    """
+    Serializer for the MappingStatus model.
+
+    Serializes the value field.
+
+    Args:
+        model (MappingStatus): The model to be serialized.
+
+        fields (list): The fields to be included in
+        the serialized output.
+    """
+
     class Meta:
         model = MappingStatus
         fields = ["value"]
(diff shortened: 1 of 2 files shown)


app/api/api/serializers.py
```diff
@@ -430,7 +557,9 @@ def validate_scan_report_file(self, value):
         wb = openpyxl.load_workbook(filename=BytesIO(file_in_memory), data_only=True)
 
         try:
-            self.run_fast_consistency_checks(wb)
+            # Store table names in the serializer
+            logger.info("Collecting table names from the scan report...")
+            self.sr_table_names = self.run_fast_consistency_checks(wb)
```
AndrewThien, line 562: line 412 will return `True, table_names` --> `self.sr_table_names` will receive the exact values (both True, and the list of table names). --> Line 274 will fail

app/api/api/serializers.py
```diff
@@ -168,18 +245,54 @@ def validate_data_dictionary_file(self, value):
...
+            )
+            logger.error(f"{missing} are missing for validation.")
+            raise ValueError(f"{missing} are required for validation.")
+
+        # Validate each csv_name against sr_table_names
+        for csv_name in csv_file_names:
+            if csv_name not in self.sr_table_names:
+                errors.append(
+                    ParseError(
+                        f"CSV name in Data Dictionary '{csv_name}' is not matching any table in Scan Report. "
```
AndrewThien, line 273:
```suggestion
                        f"'{csv_name}' in the column "csv_file_name" in the Data Dictionary does not match any table name in the Scan Report. Hint: Make sure the extension '.csv' is included"
```

app/api/api/serializers.py
```diff
@@ -168,18 +245,54 @@ def validate_data_dictionary_file(self, value):
...
+                        f"Available tables: {', '.join(sorted(self.sr_table_names))}"
```
AndrewThien, line 274:
```suggestion
                        f"Available table names/CSV file names: {', '.join(sorted(self.sr_table_names))}"
```

app/api/api/serializers.py
```diff
@@ -168,18 +245,54 @@ def validate_data_dictionary_file(self, value):
...
+                    )
+                )
+
         if errors:
             raise ParseError(errors)
 
+        logger.info("Data dictionary file is valid and ready for upload.")
+        data_dictionary.seek(0)
```
AndrewThien, line 282: Do we need this?

app/api/api/serializers.py
```diff
@@ -168,18 +245,54 @@ def validate_data_dictionary_file(self, value):
                         )
                     )
 
+            # Collect csv_file_name if line exists
+            if line and len(line) > 0:
+                csv_file_names.add(line[0].strip())
+
+        # Log the collected csv_file_names and sr_table_names
+        logger.info(f"CSV file names: {csv_file_names}")
```
AndrewThien, line 253: Logger is good for debugging, but I think we can reduce the number of loggers in this function?

app/api/api/serializers.py
Karthi-DStech: Yes. It is helpful in silent failures and resets the file pointer to the start before upload.

app/api/api/serializers.py
Karthi-DStech: Ok👍🏻

app/api/api/serializers.py
Karthi-DStech: Ok👍🏻

app/api/api/serializers.py
AndrewThien: can you double check if True is using somewhere? I think it only there to pass the try .. except block

app/api/api/serializers.py
AndrewThien: cool thanks

app/api/api/serializers.py
Karthi-DStech: I feel it is safer to be there (True) than to delete it. 

```
 _, self.sr_table_names = self.run_fast_consistency_checks(wb)
 ```
 
 I simply unpacked it to avoid the conflicts.

AndrewThien (approved):

anwarfg (approved):

brian-kim31 (approved):

Issue #1006, opened 2025-04-07
<domain>_source_concept_id and <domain>_concept_id are always the same


AndrewThien:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When generating mapping rules in Carrot, either by auto-mapping or manual mapping, the `<domain>_source_concept_id` and `<domain>_concept_id` are always the same.

### Expected Behavior

`<domain>_source_concept_id` and `<domain>_concept_id` can be the same when the coded value already has the standard concept tied. 
If not, the `<domain>_source_concept_id`  should be the concept ID of the original value/code before converting, and the ` <domain>_concept_id` should be the standard concept ID (concept_id_2 of relationship `Maps to`)

### Environment

```markdown
- OS:
- Other environment details:
```

### Are you willing to contribute to resolve this issue?

✅ Yes, me or my team intend to do the development.

PR #1019, opened 2025-04-22
Add mapping rules creation and its prepration steps to Airflow app

### LINKED ISSUE #976: Add V concepts and mapping rules creation for Airflow

### Summary

We need the new QueueMessage system and the new functions apps to work locally and on the cloud.

### Acceptance Criteria

  - [ ] New functions app works on both local and cloud
  - [ ] QueueMessage system works on both local and cloud
  - [ ] Current Azure functions of Carrot still working well
  - [ ] Q: Can API app use this to interact with current Azure functions? If so, make the brokers system configurable.
  

### Tasks

- [ ] Testing the system on local
- [ ] Deploy to a Container app
- [ ] Test if the current Carrot instance on Azure can interact with the new functions app and queueMessage system 

### Confirm creation

- [x] This issue is ready

AndrewThien:
| |
|-|
♻️ Refactor
✨ Feature

## PR Description
This PR upgraded the Airflow workers app by adding database queries and functions to find the OMOP table and fields for the standard concepts created in the `temp_standard_concepts` table. It also added the function to generate Mapping rules based on the source data and destination fields.

Functions to update job status, which are then shown in the UI, have also been added.

The workflow to publish images for Airflow containers is been modified to `merge on Master branch` to accelerate the development/testing process of the Airflow workers app.

This PR also brought back the `destination_field` into the UI of `Mapping rules` and `Review rules` pages, in favour of clarifying the difference between mapping rules (related https://github.com/Health-Informatics-UoN/carrot-mapper/discussions/958).

The logic and SQL query in this PR also fixed some issues from the original logic in workers/shared app, for auto-mapping process:
- The original logic can't generate mapping rules for standard concepts having the `Specimen` domain, even though the standard concepts were found and created. Related https://github.com/Health-Informatics-UoN/carrot-mapper/issues/406
- If one concept code is found with the vocab id provided and, even if that concept doesn't have any "Maps to" relationship, Carrot will tie the non-standard concept to the value (point to source concept instead of null) 
- Added some TODOs note for the related bug #1006

Closes #976

## ✅ Added/updated tests?
This PR needed testing on deployed instances.


app/airflow/dags/libs/core_concepts_creation.py
```diff
@@ -0,0 +1,217 @@
...
+            kwargs.get("dag_run", {}).conf.get("field_vocab_pairs")
+        )
+
+        if not field_vocab_pairs:
+            logging.error("No field-vocabulary pairs provided")
+            raise ValueError("No field-vocabulary pairs provided")
+
+        # Create the temporary table once, outside the loop, with all the columns needed
+        create_table_query = f"""
+        DROP TABLE IF EXISTS temp_standard_concepts_{table_id};
```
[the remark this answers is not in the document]
AndrewThien: This temp table is created to support the mapping rules creation process later on. So I think we can drop them after the whole process is finished and only keep the table with real, persistent data. If we clean the contents instead, the database will be full of temp_table and we may not like that :D 

For more information, the current logic is not done, and these temp tables are still needed for double-checking/testing (in both dev and deployed instances). When the logic is ready, at the end of the general workflow, they will be dropped --> this DROP TABLE command will be at the end of the workflow instead of beginning

app/airflow/dags/libs/core_concepts_creation.py
```diff
@@ -0,0 +1,217 @@
...
+            )
+            # Insert data for each field-vocabulary pair
+            find_standard_concepts_query = f"""
+            INSERT INTO temp_standard_concepts_{table_id} (sr_value_id, source_concept_id, standard_concept_id)
+            SELECT
+                srv.id AS sr_value_id,
+                c1.concept_id AS source_concept_id,
+                c2.concept_id AS standard_concept_id
+            FROM mapping_scanreportvalue srv
+            JOIN omop.concept c1 ON
```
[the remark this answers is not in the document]
AndrewThien: you mean the names of them, right?
how about: `c1 --> src_concept` and `c2 --> std_concept` ?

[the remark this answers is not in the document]
prquinlan: Yes

app/airflow/dags/libs/core_concepts_creation.py
```diff
@@ -0,0 +1,217 @@
...
+        raise
+
+
+def create_standard_concepts(**kwargs):
+    """
+    Create standard concepts for field values in the mapping_scanreportconcept table.
+    Only inserts concepts that don't already exist.
+    """
+    try:
+        scan_report_id = kwargs.get("dag_run", {}).conf.get("scan_report_id")
```
prquinlan, line 129: Can these be renamed, to avoid this further down: 
scan_report_id=scan_report_id,
table_id=table_id,

This might be my lack of python knowledge, so ignore if I am being silly

app/airflow/dags/libs/core_concepts_creation.py
```diff
@@ -0,0 +1,217 @@
...
+
+        if not table_id:
+            logging.warning("No table_id provided in find_sr_concept_id")
+
+        update_query = f"""        
+        -- Update sr_concept_id with the mapping_scanreportconcept ID
+        UPDATE temp_standard_concepts_{table_id} tsc
+        SET sr_concept_id = src.id
+        FROM mapping_scanreportconcept src
+        WHERE src.object_id = tsc.sr_value_id
```
prquinlan, line 202: tsc and src

app/airflow/dags/libs/core_prep_rules_creation.py
```diff
@@ -0,0 +1,346 @@
...
+    two main SQL operations:
+
+    1. Maps concept domains to appropriate OMOP tables (e.g., 'Race'/'Gender'/'Ethnicity'
+       to 'person', 'Condition' to 'condition_occurrence', etc.) and sets the dest_table_id
+       in the temporary concepts table.
+    2. Adds the person_id field reference to each record by finding the corresponding field
+       in the mapping_omopfield table.
+    """
+    try:
+        table_id = kwargs.get("dag_run", {}).conf.get("table_id")
```
prquinlan, line 28: same as before with table_id and then in line 31

app/airflow/dags/libs/core_prep_rules_creation.py
```diff
@@ -0,0 +1,346 @@
...
+            logging.warning(
+                "No table_id provided in find_dest_table_and_person_field_id"
+            )
+
+        core_query = f"""
+        -- Update destination table ID
+        UPDATE temp_standard_concepts_{table_id} tsc
+        SET dest_table_id = ot.id
+        FROM omop.concept std_concept
+        LEFT JOIN mapping_omoptable ot ON
```
prquinlan, line 46: ot and tsc

prquinlan (commented):
Another general comment on the SQL, if you could swap things like this:
        UPDATE temp_standard_concepts_{table_id} tsc

to be: 
        UPDATE temp_standard_concepts_{table_id} as tsc

It is optional in Postgres, but I didnt realise that, I think best practice would be to include it.

app/airflow/dags/libs/core_prep_rules_creation.py
```diff
@@ -166,82 +165,62 @@ def find_concept_fields(**kwargs):
...
-            source_concept_field_id = dfm.source_concept_field_id,
-            source_value_field_id = dfm.source_value_field_id,
-            dest_concept_field_id = dfm.dest_concept_field_id
-        FROM domain_field_map dfm
-        WHERE dfm.sr_value_id = tsc.sr_value_id;
-        """
+        # Create a staging table with computed field values
+        stage_query = f"""
+        -- Create a staging table with the correct field IDs for each record
+        CREATE TABLE temp_concept_fields_staging_{table_id} AS
```
prquinlan, line 179: Is this missing an alias name, nothing after As

app/airflow/dags/libs/core_concepts_creation.py
```diff
@@ -0,0 +1,217 @@
...
+                concept_id,
+                content_type_id
+            )
+            SELECT
+                NOW(),
+                NOW(),
+                tsc.sr_value_id,
+                'V', -- Creation type: Built from Vocab dict
+                tsc.standard_concept_id,
+                23 -- content_type_id for scanreportvalue
```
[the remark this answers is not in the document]
AndrewThien: yeah, I would need alias here for the checking of existing concepts at lines 156-157. So I will improve the name. tsc --> `temp_std_concepts` is okay?

app/airflow/dags/libs/core_concepts_creation.py
AndrewThien: I updated the parameter names of the function `update_job_status` instead. I think this can also avoid the confusion may caused

app/airflow/dags/libs/core_prep_rules_creation.py
AndrewThien: same with the reply for `scan_report_id=scan_report_id,`

app/airflow/dags/libs/core_prep_rules_creation.py
AndrewThien: as explained in DM: this AS here is for the command CREATE TABLE ... AS SELECT .... which creates a new table using the results of a SELECT statement.

app/airflow/dags/libs/core_concepts_creation.py
```diff
@@ -0,0 +1,217 @@
...
+            )
+            # Insert data for each field-vocabulary pair
+            find_standard_concepts_query = f"""
+            INSERT INTO temp_standard_concepts_{table_id} (sr_value_id, source_concept_id, standard_concept_id)
+            SELECT
+                sr_value.id AS sr_value_id,
+                src_concept.concept_id AS source_concept_id,
+                std_concept.concept_id AS standard_concept_id
+            FROM mapping_scanreportvalue sr_value
+            JOIN omop.concept src_concept ON
```
prquinlan, line 90: Sorry to be a pain

JOIN omop.concept As src_concept ON

prquinlan: Can you check other joins too

prquinlan (dismissed):

app/airflow/dags/auto_mapping_orchestrator_dag.py
Karthi-DStech, line 1: Can we add a docstring on top of the file for better documentation purposes? 

for eg: 

"""
DAG: auto_mapping_orchestrator

Triggers the create_concepts_from_vocab_dict DAG for automatic creation of OMOP mapping rules.

Requires table_id, person_id_field, date_event_field, scan_report_id, and field_vocab_pairs in conf.
"""

app/airflow/dags/concepts_from_vocab_dag.py
```diff
@@ -1,8 +1,24 @@
...
+    find_dest_table_and_person_field_id,
+    find_date_fields,
+    find_concept_fields,
+    find_additional_fields,
+)
+
+from libs.core_rules_creation import (
+    delete_mapping_rules,
+    create_mapping_rules,
+)
```
Karthi-DStech, line 22: Can we add a file-level docstring at the top of this DAG for documentation and onboarding? This should briefly describe what the DAG does and what configuration keys are required in the conf.

app/airflow/dags/concepts_from_vocab_dag.py
```diff
@@ -17,27 +33,104 @@
...
+find_std_concepts_task = PythonOperator(
+    task_id="find_standard_concepts",
+    python_callable=find_standard_concepts,
+    provide_context=True,
+    dag=dag,
+)
+
+
+create_standard_concepts_task = PythonOperator(
+    task_id="create_standard_concepts",
```
Karthi-DStech, line 66: We can refactor the repeated PythonOperator instantiations into a reusable component for cleaner code and follow DRY (Don't Repeat Yourself) code. 

For example, by defining a small helper function, we reduce duplication and make adding or modifying tasks easier in the future:

```
def _create_task(task_id, python_callable, provide_context=True):
    return PythonOperator(
        task_id=task_id,
        python_callable=python_callable,
        provide_context=provide_context,
        dag=dag,
    )

find_std_concepts_task = _create_task(task_id="find_standard_concepts", python_callable=find_standard_concepts)
```

app/airflow/dags/concepts_from_vocab_dag.py
```diff
@@ -17,27 +33,104 @@
...
     dag=dag,
 )
-# TODO: add tasks to do error handling and update status
-# TODO: create workflow - deploy MVP - testing
 
 # End the workflow
 end = EmptyOperator(task_id="end", dag=dag)
 
-(start >> core_task >> end)
+(
```
Karthi-DStech, line 124: I assume that all tasks are dependent on each other and must run sequentially. 

If it was true, then can we can simplify the code by putting the tasks in a list and chaining them inside a loop like this:

```
tasks = [
    find_std_concepts_task,
    create_standard_concepts_task,
    # ... other tasks
]

curr = start
for task in tasks:
    curr >> task 
    curr = task
curr >> end  

```
In this way, each task waits for the previous one to complete and the last task is linked to the 'end' operator as well. 

Let me know your thoughts.

app/airflow/dags/auto_mapping_orchestrator_dag.py
Karthi-DStech, line 1: Why didn't we add retries now? Are we going to add retries later, mate?

AndrewThien: I purposely didn't add retries because I want the DAGs to fail fast (if they want to fail) and don't need to wait to see the error. That's for dev stage, and of course in prod I will add retries

app/airflow/dags/concepts_from_vocab_dag.py
AndrewThien: Does your suggestion do the same thing, and we also have to make sure the order in the `tasks` list is correct? Sorry, I couldn't find the added improvement from the `for` loop here.

app/airflow/dags/libs/core_concepts_creation.py
```diff
@@ -0,0 +1,217 @@
...
+def find_standard_concepts(**kwargs):
+    """
+    Maps source field values to standard OMOP concepts based on field-vocabulary pairs.
+
+    Creates a temporary table to store mapping information between source values and
+    standard concepts. For each field-vocabulary pair provided, finds corresponding
+    standard concepts in the OMOP vocabulary and inserts them into the temporary table.
+
+    """
+    try:
```
Karthi-DStech, line 23: Assuming table_id is an integer, can we add a validator before interpolating table_id into SQL queries?

```
if not isinstance(table_id, int):
    raise ValueError(f"Invalid table_id: {table_id}. Must be an integer.")
```

Let me know your thoughts.

app/airflow/dags/libs/core_concepts_creation.py
```diff
@@ -0,0 +1,217 @@
...
+)
+from airflow.providers.postgres.hooks.postgres import PostgresHook
+import logging
+
+# PostgreSQL connection hook
+pg_hook = PostgresHook(postgres_conn_id="postgres_db_conn")
+
+
+def find_standard_concepts(**kwargs):
+    """
```
Karthi-DStech, line 15: For each function that uses **kwargs, can we add to the docstring which keys are expected? 

This will improve readability and make it easier for others to understand how to use the function correctly, including which parameters are required.

app/airflow/dags/libs/core_concepts_creation.py
```diff
@@ -0,0 +1,217 @@
...
+        logging.error(f"Error in find_standard_concepts: {str(e)}")
+        raise
+
+
+def create_standard_concepts(**kwargs):
+    """
+    Create standard concepts for field values in the mapping_scanreportconcept table.
+    Only inserts concepts that don't already exist.
+    """
+    try:
```
Karthi-DStech, line 128: If table_id is an integer, can we add a validator before interpolating table_id into SQL queries? 

```
if not isinstance(table_id, int):
    raise ValueError(f"Invalid table_id: {table_id}. Must be an integer.")
```
Let's make sure `table_id` is present and is an integer before using it in string-formatted SQL, to avoid any accidental injection or run-time errors.

app/airflow/dags/libs/core_concepts_creation.py
AndrewThien: this table_id will be provided from backend of Carrot (so if something is wrong, it must be wrong from steps/process before this Airflow passing. And I think we already have validator for it). Also, I believe, if we pass JSON data in a POST request, all of the data will be `stringify`-ed ? Anyway, I tested with int and string types, and Airflow is running fine. 

Or do you mean we should convert the variable to `number` and then check if it is not a float?

app/airflow/dags/libs/core_concepts_creation.py
```diff
@@ -0,0 +1,217 @@
...
+    StageStatusType,
+)
+from airflow.providers.postgres.hooks.postgres import PostgresHook
+import logging
+
+# PostgreSQL connection hook
+pg_hook = PostgresHook(postgres_conn_id="postgres_db_conn")
+
+
+def find_standard_concepts(**kwargs):
```
Karthi-DStech, line 14: Can we consider adding type hints for functions that use **kwargs?? 

For Example:

```
def find_standard_concepts(**kwargs: dict) -> None:
```

app/airflow/dags/libs/core_concepts_creation.py
```diff
@@ -0,0 +1,217 @@
...
+                stage=JobStageType.BUILD_CONCEPTS_FROM_DICT,
+                status=StageStatusType.FAILED,
+                details=f"Error in create_standard_concepts: {str(e)}",
+            )
+    except Exception as e:
+        logging.error(f"Error in create_standard_concepts: {str(e)}")
+        raise
+
+
+def find_sr_concept_id(**kwargs):
```
Karthi-DStech, line 185: Can we consider adding type hints for functions that use **kwargs?? 

For Example:

```
def find_sr_concept_id(**kwargs: dict) -> None:
```

app/airflow/dags/libs/core_concepts_creation.py
```diff
@@ -0,0 +1,217 @@
...
+    """
+    Update temp_standard_concepts table with sr_concept_id column
+    containing the IDs of standard concepts added to mapping_scanreportconcept.
+    This will help the next steps to be shorter.
+    """
+    try:
```
Karthi-DStech, line 191: Assuming table_id is an integer, can we add a validator before interpolating table_id into SQL queries?

```
if not isinstance(table_id, int):
    raise ValueError(f"Invalid table_id: {table_id}. Must be an integer.")
```

Let me know your thoughts.

app/airflow/dags/libs/core_concepts_creation.py
```diff
@@ -0,0 +1,217 @@
...
+            )
+    except Exception as e:
+        logging.error(f"Error in create_standard_concepts: {str(e)}")
+        raise
+
+
+def find_sr_concept_id(**kwargs):
+    """
+    Update temp_standard_concepts table with sr_concept_id column
+    containing the IDs of standard concepts added to mapping_scanreportconcept.
```
Karthi-DStech, line 188: For each function that uses **kwargs, can we add to the docstring which keys are expected? 

This will improve readability and make it easier for others to understand how to use the function correctly, including which parameters are required.

example: 
```
Expects in kwargs['dag_run'].conf:
        - scan_report_id (int)
        - table_id (int)
        - field_vocab_pairs (list/dict)
```

app/airflow/dags/libs/core_concepts_creation.py
```diff
@@ -0,0 +1,217 @@
...
+                    details=f"Error in find_standard_concepts_query: {str(e)}",
+                )
+                raise
+    except Exception as e:
+        logging.error(f"Error in find_standard_concepts: {str(e)}")
+        raise
+
+
+def create_standard_concepts(**kwargs):
+    """
```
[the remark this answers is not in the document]
AndrewThien: In the `auto_mapping_orchestrator`, at the triggering DAG step, we will provide all of the configs to `create_concepts_from_vocab_dict` and then these configs will be passed to the children functions by `provide_context=True`. I didn't find the way to pass the config for each child DAG separately, and that's why all of the children DAGs will have the same config, or they will expect the same thing --> Do we really need to docstring this?

app/airflow/dags/libs/core_concepts_creation.py
AndrewThien: will the same reply above help here as well?

app/airflow/dags/libs/core_concepts_creation.py
AndrewThien: hope the explaination below help?

app/airflow/dags/auto_mapping_orchestrator_dag.py
AndrewThien: okay

app/airflow/dags/concepts_from_vocab_dag.py
AndrewThien: okay, that's good idea, let me try it

app/airflow/dags/concepts_from_vocab_dag.py
Karthi-DStech: Yes, the benefit of using a list and a loop to build the chain is mainly about readability and ease of future changes:

```
start 
>> find_std_concepts_task 
>> create_standard_concepts_task 
>> ... 
>> end
```

We could try to do something like this if it works without any error

```
tasks = [
    _create_task("find_standard_concepts", find_standard_concepts),
    _create_task("create_standard_concepts", create_standard_concepts),
    _create_task("find_sr_concept_id", find_sr_concept_id),
    _create_task("find_dest_table_and_person_field_id", find_dest_table_and_person_field_id),
    _create_task("find_date_fields", find_date_fields),
    _create_task("find_concept_fields", find_concept_fields),
    _create_task("find_additional_fields", find_additional_fields),
    _create_task("delete_mapping_rules", delete_mapping_rules),
    _create_task("create_mapping_rules", create_mapping_rules),
]

curr = start
for task in tasks:
    curr >> task
    curr = task
curr >> end
```

The loop approach is more maintainable and tidy for sequential workflows, especially as the number of steps grows.

Let me know if you’d like a more detailed explanation or if you’re happy with this one.

app/airflow/dags/auto_mapping_orchestrator_dag.py
Karthi-DStech: Make sense. Thanks for the reply.

app/airflow/dags/libs/core_concepts_creation.py
Karthi-DStech: Thanks for the details, Thien. You're right that if table_id is already validated upstream and always passed as an integer, that's ideal and we can skip it. 

Yes, converting to int after validating it is all digits (and rejecting floats or anything else unexpected) would be sufficient protection in case the value is stringified or coming from any external system, now or in the future.

Even if Airflow works when passing as a string, protecting the SQL layer is an extra safety net.

example: 
```
if not str(table_id).isdigit():
    raise ValueError(f"Invalid table_id: {table_id}")
table_id = int(table_id)
```

Either we could consider doing this or ignore this because we are already validating that. What we could do?

app/airflow/dags/concepts_from_vocab_dag.py
AndrewThien: yeah, the only save here is the `>>` symbol, but yeah, seems to be a good save. Let me try

app/airflow/dags/libs/core_concepts_creation.py
Karthi-DStech: I thought that explicitly documenting the expected keys in the docstring can be useful, especially for anyone new working on this code or returning to it later.

While the code is perfect (the code itself makes it clear what’s needed), having these listed up front means people don’t have to read through the logic to figure out exactly what config parameters they need to supply (and in what structure), making the code more self-documenting.

It’s not strictly required, but it will improve code readability, follow best practices and maintenance for our team.

Let me know if you think this helps.

- If yes, we can add these docstring details.
- If not, no worries—I can remove my related comments.

app/airflow/dags/libs/core_concepts_creation.py
Karthi-DStech: Thanks, that's a good suggestion. Sorry for the inconvenience. I am noting down everything separately and pasting it.

app/airflow/dags/concepts_from_vocab_dag.py
Karthi-DStech: Sure. Let me know if I could do anything from my side.

app/airflow/dags/libs/core_concepts_creation.py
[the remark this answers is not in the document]
Karthi-DStech: Ah, I get it now. Maybe we could add for better documentation or we can ignore it. What do you feel about this mate?

app/airflow/dags/libs/core_concepts_creation.py
AndrewThien: okay, technically it is not necessary because I couldn't find any reasons for this ID to suddenly become float, for example, but I will add some checks and conversion here. Doing defensive safety costs not much :D

Karthi-DStech: That's fine then. Then we could ignore it.

app/airflow/dags/concepts_from_vocab_dag.py
AndrewThien: This worked so well. Thannks a lot

Karthi-DStech: Glad to help😄

AndrewThien:
Hi @Karthi-DStech,
I believe all of your comments (which are really helpful :D) have been addressed via pushed commits. Please have a re-review when you can. Thanks a lot!

Karthi-DStech:
> Hi @Karthi-DStech, I believe all of your comments (which are really helpful :D) have been addressed via pushed commits. Please have a re-review when you can. Note: The logic now has changed a bit: all of the params in the task instance will be typed, validated and converted in the first place, then pulled from xcom and used downstream, instead of being taken raw from the dag_run config, then validating + converting. Thanks a lot!

Greetings @AndrewThien, 

I hope you’re doing well. Thank you for your prompt response and for addressing all my comments so thoroughly. I’m glad my suggestions were helpful and always happy to contribute.

I have reviewed the updated code, and everything looks great to me. I’m happy to approve these changes and thank you for your excellent work!

Karthi-DStech (approved):

AndyRae (approved):

PR #1032, opened 2025-05-02
fix #1031

### LINKED ISSUE #1031: CORS_ALLOW_CREDENTIALS needs to be a boolean

### Current Behavior

When installing the system from scratch, an error occurs during setup preventing use of the system because the `CORS_ALLOW_CREDENTIALS` setting is set to a string `"False"` rather than the boolean `False`

### Expected Behavior

Accounts should be created as usual and the 

### Steps To Reproduce

1. create an instance from scratch including (importantly) creating the databases
2. observe that the admin panel shows a Django error related to missing tables
      - the tables were not created because the installation logic crashed when `CORS_ALLOW_CREDENTIALS` was read as a string
4. observe that one cannot login to the system

### Environment

```markdown
- OS:
- Other environment details:
```

_No response_

### Anything else?

Changing this line ...

    https://github.com/Health-Informatics-UoN/carrot-mapper/blob/master/app/api/config/settings.py#L97

... to this will fixit ...

```python
CORS_ALLOWED_CREDENTIALS = os.getenv("CORS_ALLOW_CREDENTIALS", str(False)).lower() in ("true", "1")
```

None

g-pechorin:
| |
|-|
🦋 Bug Fix

## PR Description

Fixes an issue with the configuration that caused a problem during install.

Related #1031
Closes #1031

## ✅ Added/updated tests?

## [optional] What gif best describes this PR or how it makes you feel?
this is a tedious detail - sorry

### DIFF
--- a/app/api/config/settings.py
+++ b/app/api/config/settings.py
@@ -94,7 +94,7 @@ MIDDLEWARE = [
     "whitenoise.middleware.WhiteNoiseMiddleware",
 ]
 
-CORS_ALLOWED_CREDENTIALS = os.getenv("CORS_ALLOW_CREDENTIALS", str(False)).lower() in (
+CORS_ALLOW_CREDENTIALS = os.getenv("CORS_ALLOW_CREDENTIALS", "False").lower() in (
     "true",
     "1",
 )
--- a/app/api/config/settings.py
+++ b/app/api/config/settings.py
@@ -94,7 +94,10 @@ MIDDLEWARE = [
     "whitenoise.middleware.WhiteNoiseMiddleware",
 ]
 
-CORS_ALLOW_CREDENTIALS = os.getenv("CORS_ALLOW_CREDENTIALS", "False")
+CORS_ALLOWED_CREDENTIALS = os.getenv("CORS_ALLOW_CREDENTIALS", str(False)).lower() in (
+    "true",
+    "1",
+)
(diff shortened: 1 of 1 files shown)


app/api/config/settings.py
```diff
@@ -94,7 +94,10 @@
     "whitenoise.middleware.WhiteNoiseMiddleware",
 ]
 
-CORS_ALLOW_CREDENTIALS = os.getenv("CORS_ALLOW_CREDENTIALS", "False")
+CORS_ALLOWED_CREDENTIALS = os.getenv("CORS_ALLOW_CREDENTIALS", str(False)).lower() in (
+    "true",
+    "1",
+)
```
AndyRae, line 100:
```suggestion
CORS_ALLOW_CREDENTIALS = os.getenv("CORS_ALLOW_CREDENTIALS", "False").lower() in (
    "true",
    "1",
)
```

AndyRae: I couldn't find any docs for `ALLOWED`, unless I've misunderstood?

AndyRae (changes requested):

app/api/config/settings.py
g-pechorin: as is - the mapper doesn't install without these changes

i'm testing this other change (now)

brian-kim31 (approved):

Karthi-DStech (approved):

AndyRae (approved):

Issue #1036, opened 2025-05-06
Restructuring Docker compose and Enviroment variables


AndrewThien:
### Is this the right issue type?

- [x] Yes, I'm planning work for this project team.

### Summary

Since we will have a new `workers` solution (i.e., Airlfow) alongside Azure functions, making a complete cloud-agnostic solution when combining with Minio, a restructuring of Docker Compose stacks is required.

Note: This piece of work is for Developers in the dev environment. In deployed environments, we need to make sure crossing use of these services can happen. For example, Airflow can interact with both Azurite and Minio.

Proposed steps:

1. Create a Docker Compose profile named [cloud-agnostic] (or something similar) and put the services `Minio`, `airflow-webserver` and `scheduler` here. --> Change `minio` and `airflow` type in the code to `cloud-agnostic`
2. Create a Docker Compose profile named [azure] and add the services `workers` (the name needs to be changed) and `azurite` here.
3. For quickstart purposes, the user only needs to run one command `docker compose up -d`, and Carrot can be up and running locally AND efficiently.
 4. Provide a second option to run Docker Compose:
      - If choose `azure` as default in step 3: `WORKERS_SOLUTION=cloud-agnostic docker compose --profile cloud-agnostic up -d`
      - If choose `cloud-agnostic` as default in step 3: `WORKERS_SOLUTION=azure docker compose --profile azure up -d`

### Acceptance Criteria

- [ ] Criterion A is met.
- [ ] Criterion B is met.
- [ ] Tests pass.
- [ ] Documentation added.

### Tasks

- [ ] Task A.
- [ ] Task B.
- [ ] Unit tests.
- [ ] Documentation.

### Confirm creation

- [x] This issue is ready

PR #1037, opened 2025-05-06
Add Carrot-Airflow connection

### LINKED ISSUE #1035: Airflow app part 5: Airflow and Carrot Mapper integration

### Summary

After each feature in Airflow is created, we need a "connection" or trigger point for Carrot's backend to be able to send request to Airlfow.

### Acceptance Criteria

- [ ] Carrot can request Auto-mapping process features from the Azure function or Airflow
- [ ] Carrot can request the uploaded Scan report process features from the Azure function or Airflow
- [ ] Carrot can request the process to create the mapping files features from Azure function or Airflow

### Tasks

- [ ] Task A.
- [ ] Task B.
- [ ] Unit tests.
- [ ] Documentation.

### Confirm creation

- [x] This issue is ready

AndrewThien:
| |
|-|
♻️ Refactor
✨ Feature

## PR Description
This PR added the triggering endpoints for Carrot to interact with Airflow using API calls. Since now on, the interactions between Backend and Workers (either Azure functions or Airflow) will determine and execute in the `FunctionService`.

This PR also solve the issue #1039 in Airflow app.

Note: 
- Naming of types/classes can be improved when we tackle the issue #1036   
- Function `trigger_scan_report_processing` and `trigger_rules_export` for Airflow workers are using the methods from `Azure` for now. This helps devs reduce time to change the environment. vars

Closes #1035 

## ✅ Added/updated tests?
All of triggers from BE to Azure functions are working normally


app/airflow/dags/libs/find_standard_V_concepts.py
```diff
@@ -111,6 +137,8 @@ def find_standard_concepts(**kwargs) -> None:
                         "table_id": table_id,
                         "sr_field_id": sr_field_id,
                         "vocabulary_id": vocabulary_id,
+                        "skip_domains_check": skip_domains_check,
```
Karthi-DStech, line 140: Are we skipping the domain checks here? If yes, may I know why?

app/api/api/views.py
```diff
@@ -399,8 +395,8 @@ def perform_create(self, serializer, non_file_serializer):
                 use_read_method=False,
             )
 
-        # send to the upload queue
-        add_message(os.environ.get("WORKERS_UPLOAD_NAME"), azure_dict)
+        # send to the workers service
+        function_service.trigger_scan_report_processing(azure_dict)
```
Karthi-DStech, line 399: I assume we are now using a common model to process the scan report. If that's the case, can we change the name from `azure_dict` to something else?

app/shared/shared/services/function_service.py
```diff
@@ -0,0 +1,202 @@
...
+    """
+
+    def __init__(self):
+        """
+        Initialize the FunctionService with configuration from Django settings.
+
+        Sets up the service with the configured function type (Azure or Airflow)
+        and the base URL for Airflow API calls if applicable.
+        """
+        self._function_type = settings.FUNCTION_TYPE
```
Karthi-DStech, line 39: Can we get the function type from the `os.getenv`?

app/shared/shared/services/function_service.py
```diff
@@ -0,0 +1,202 @@
...
+            data_dictionary_name: Optional name of a data dictionary blob to use
+                                 for field vocabulary mappings.
+            trigger_reuse_concepts: Flag to enable/disable the reuse of previously
+                                   mapped concepts (default: True).
+
+        Raises:
+            ValueError: If the configured function type is not supported.
+            HTTPError: If the HTTP request to trigger the process fails.
+        """
+        self._validate_function_type()
```
Karthi-DStech, line 115: Can we initialise this validation globally in the `__init__` method to avoid duplicate checks across different functions? 

def __init__(self):
     self._validate_function_type()
     
The same applies to other places, mate.

app/shared/shared/services/function_service.py
```diff
@@ -0,0 +1,202 @@
...
+                json=body,
+                auth=HTTPBasicAuth(
+                    settings.AIRFLOW_ADMIN_USERNAME, settings.AIRFLOW_ADMIN_PASSWORD
+                ),
+            )
+            response.raise_for_status()
+            return response
+        except requests.exceptions.HTTPError as e:
+            logging.error(f"HTTP Trigger failed: {e}")
+            raise
```
Karthi-DStech, line 86: Why only raise? Are we trying to capture the error message and fetch it in the UI later?

Karthi-DStech (changes requested):
Looks good to me, Thien. Great Job! Just slight suggestions for improvements and other than that, I have no problem merging it into the main branch.

app/shared/shared/services/function_service.py
```diff
@@ -0,0 +1,202 @@
...
+from django.conf import settings
+from shared.services.azurequeue import add_message
+
+
+class FUNCTION_TYPE(StrEnum):
+    AZURE = "azure"
+    AIRFLOW = "airflow"
+
+
+class FunctionService:
```
AndyRae, line 19: `FunctionService` seems too tightly coupled to describe Azure Functions - given the rest of the codebase effectively calls them Jobs, maybe just `JobService`?

Or `BackgroundJobService`

app/shared/shared/services/function_service.py
AndyRae: The preferred pattern for all settings to be configured in `settings.py`, which reads from the environment. Rather than scattering lots of os.getenv across the codebase.

app/api/config/settings.py
```diff
@@ -264,6 +266,9 @@
 # Storage Type Variable
 STORAGE_TYPE = os.getenv("STORAGE_TYPE", "minio")
 
+# Function Type Variable
+FUNCTION_TYPE = os.getenv("FUNCTION_TYPE", "azure")
```
AndyRae, line 270: Similarly on naming - `BACKGROUND_JOB_TYPE` might be better?

app/shared/shared/services/function_service.py
```diff
@@ -0,0 +1,202 @@
...
+    This service abstracts the implementation details of triggering data processing
+    workflows in either Azure Functions or Apache Airflow environments. It provides
+    a consistent interface for initiating common operations like auto-mapping,
+    scan report processing, and rules export regardless of the underlying execution platform.
+
+    The service determines which platform to use based on the FUNCTION_TYPE setting
+    and handles the appropriate API calls or message queue operations accordingly.
+    """
+
+    def __init__(self):
```
AndyRae, line 32: As a general implementation piece - I think it's worth walking through alternate approaches to what we're doing.

This is effectively dependency injection, so it would be nice to have a pattern that supports it.

I'm thinking encapsulation, testing, flexibility, and less conditional logic basically. An abstract base class of `BackgroundJobService` defining the service, which is implemented by Azure and Airflow services. 

The current approach is pragmatic - which is fine, but improved patterns for testing in particular would be good.

app/airflow/dags/libs/find_standard_V_concepts.py
AndrewThien: we are doing the `domain_check` by line 49 `skip_domains_check = False`
I want to do double negative here because in the future, if a variable, for example, `table.death_table = true` --> `skip_domains_check`, I don't need to switch anything and pass directly to the `skip_domains_check`

app/shared/shared/services/function_service.py
AndrewThien: I was thinking about `WorkerService` and wanted to do it later in #1036 , but yeah, we can do it now.

Is `WorkerService` okay?

app/shared/shared/services/function_service.py
AndrewThien: okay, let me improve it

app/shared/shared/services/function_service.py
AndrewThien: cool, I will do the validation of function_type in _init_

app/airflow/dags/libs/find_standard_V_concepts.py
Karthi-DStech: Thanks for the response, Thien. I'll solve it.

app/shared/shared/services/worker_service.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def trigger_rules_export(self, message_body: Dict[str, Any]):
+        # For now using the queue approach until Airflow implementation is completed
+        add_message(self._workers_rules_export_name, message_body)
+        # Future implementation:
+        # self._trigger_airflow_dag(self._rules_export_dag_id, message_body)
+
+
+# Function to create the appropriate service based on configuration
+def create_worker_service() -> WorkerService:
```
AndyRae, line 163: `get_worker_service` is more conventional here

app/api/config/settings.py
```diff
@@ -264,6 +266,9 @@
 # Storage Type Variable
 STORAGE_TYPE = os.getenv("STORAGE_TYPE", "minio")
 
+# Worker Service Type Variable
+WORKER_SERVICE_TYPE = os.getenv("WORKER_SERVICE_TYPE", "azure")
```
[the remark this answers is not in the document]
AndyRae: Okay I'm going to note that my comment was before you made the commit to implement this.. I just hadn't submitted the review! Nice work

Karthi-DStech (approved):
Nice work, Thien. I am happy to approve and thanks for the PR.

AndyRae (approved):

PR #1106, opened 2025-06-09
Add migration to create indexes in the DB on startup

### LINKED ISSUE #1091: Add `CREATE INDEX` on the DB for crucial tables/columns

### Summary

Currently, and in the future, workers' service's interaction with PostgreSQL DB would benefit a lot if we can create index tables for important and frequent-used tables/columns in the Database.

Note:
- The `CREATE INDEX` commands should be added to the code in the `models.py` and run through migrations.
- If the index already existed, the index creation will be skipped
- Only crucial tables/columns which have been used a lot will have their INDEX table. 

### Acceptance Criteria

- [ ] Criterion A is met.
- [ ] Criterion B is met.
- [ ] Tests pass.
- [ ] Documentation added.

### Tasks

- [ ] Task A.
- [ ] Task B.
- [ ] Unit tests.
- [ ] Documentation.

### Confirm creation

- [x] This issue is ready

AndrewThien:
| |
|-|
⚡️ Optimization

## PR Description
This PR added SQL commands to create useful indexes to spin up the process of DB querying.  

Closes #1091


AndyRae:
I would suggest we create the indexes on the models in `models.py`, and make a migration to be added, instead of running this manually. So the indexes are managed in code.

PR #1131, opened 2025-06-26
Rename public to shared

### LINKED ISSUE #1070: Scan report "public" is misleading

### Summary

The language around "PUBLIC" throughout the app is misleading, and should consider alternatives wording - replaced with `shared` ?

Image

### Acceptance Criteria

- [ ] Replace `public` with `shared`
- [ ] Tests pass.
- [ ] Documentation added.

### Tasks

- [ ] Replace `public` with `shared`
- [ ] Documentation.

### Confirm creation

- [x] This issue is ready

brian-kim31:
♻️ Refactor

## PR Description
This PR updates misleading terminology across the backend, frontend, and related files. The term "public" was previously used in ways that could confuse users regarding the visibility or sharing model of data. To improve clarity, all instances of "public" have been replaced with "shared", and unnecessary capitalization (e.g., "PUBLIC") has been removed from the UI.

Related #1070
Closes #1070

Screenshot 2025-06-26 at 17 31 10
Screenshot 2025-06-26 at 17 32 44
Screenshot 2025-06-26 at 17 32 56
Screenshot 2025-06-26 at 17 33 28

## ✅ Added/updated tests?
Screenshot 2025-06-26 at 17 08 39

### DIFF
--- a/app/shared/shared/mapping/migrations/0010_alter_dataset_visibility_alter_scanreport_visibility.py
+++ b/app/shared/shared/mapping/migrations/0010_alter_dataset_visibility_alter_scanreport_visibility.py
@@ -6,18 +6,26 @@ from django.db import migrations, models
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('mapping', '0009_delete_classificationsystem_delete_nlpmodel_and_more'),
+        ("mapping", "0009_delete_classificationsystem_delete_nlpmodel_and_more"),
     ]
 
     operations = [
         migrations.AlterField(
-            model_name='dataset',
-            name='visibility',
-            field=models.CharField(choices=[('SHARED', 'Shared'), ('RESTRICTED', 'Restricted')], default='SHARED', max_length=10),
+            model_name="dataset",
+            name="visibility",
+            field=models.CharField(
+                choices=[("SHARED", "Shared"), ("RESTRICTED", "Restricted")],
+                default="SHARED",
+                max_length=10,
+            ),
         ),
         migrations.AlterField(
-            model_name='scanreport',
-            name='visibility',
-            field=models.CharField(choices=[('SHARED', 'Shared'), ('RESTRICTED', 'Restricted')], default='SHARED', max_length=10),
+            model_name="scanreport",
+            name="visibility",
+            field=models.CharField(
+                choices=[("SHARED", "Shared"), ("RESTRICTED", "Restricted")],
+                default="SHARED",
+                max_length=10,
+            ),
         ),
     ]


app/shared/shared/mapping/migrations/0010_alter_dataset_visibility_alter_scanreport_visibility.py
AndrewThien, line 1: This migration makes me nervous because I am not sure if the current SRs, which have "Public" as visibility, will automatically have "Shared" as visibility when we apply this migration. In other words, if I am not wrong, we need to migrate the current visibility manually as well.

app/next-client-app/components/scanreports/CreateScanReportForm.tsx
```diff
@@ -189,23 +189,23 @@ export function CreateScanReportForm({
...
                       }
                     });
-                    setPublicVisibility(checked);
+                    setSharedVisibility(checked);
                   }}
                   defaultChecked
                 />
                 <Label className="text-lg">
-                  {values.visibility === "PUBLIC" ? "PUBLIC" : "RESTRICTED"}
+                  {values.visibility === "SHARED" ? "SHARED" : "RESTRICTED"}
```
AndrewThien, line 200: If we are removing the capitalisation of PUBLIC, why do we have SHARED capitalisated here?

brian-kim31: IS one of the aims removing the capitalisation? According to the issue, changing capitalisation was not specified.

app/shared/shared/mapping/migrations/0010_alter_dataset_visibility_alter_scanreport_visibility.py
brian-kim31: You have a point here

brian-kim31:
> > > a general comment/question: do we really need to change everything from `Public` to `Shared`, or do we only need to change it on the UI?
> > 
> > 
> > My thoughts were that this would be confusing to other devs why we have public in the backend and shared in the frontend, also the issue mentioned about tests passing hence why i also saw the need to change the backend
> 
> Okay, I don't mind either way tbh, just hope that nothing will break :D

Or we can do this, Let me change the frontend part only for now, then merge that. Its less risky. @AndrewThien @AndyRae

AndyRae:
> > a general comment/question: do we really need to change everything from `Public` to `Shared`, or do we only need to change it on the UI?
> 
> My thoughts were that this would be confusing to other devs why we have public in the backend and shared in the frontend, also the issue mentioned about tests passing hence why i also saw the need to change the backend

I see your point - but feel like this would be solved through documentation. For this issue - we should only change the frontend. It's very normal to have the UI representation not match the API, it's why we have the separation of concerns to enable this, and enables us to move faster.

PR #1141, opened 2025-07-04
Fix Dependency error upon merge

brian-kim31:
🦋 Bug Fix

## PR Description
This PR fixes dependency review error when jobs are runnning upon merge

Screenshot 2025-07-04 at 13 57 54

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -205,6 +205,7 @@ jobs:
   # Deploy Workers Dev
   deploy-workers-dev:
     runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
     needs: [build-and-publish-workers, build-and-publish-web]
     environment: dev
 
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -21,6 +21,7 @@ permissions:
 
 jobs:
   dependency-review:
+    if: github.event_name == 'pull_request' || github.event_name == 'pull_request_target' || github.event_name == 'merge_group'
```
AndyRae, line 24: Sorry hadn't spotted this in the original PR - the `dependency-review` action shouldn't be in the `deploy.yml` workflow, it should be it's own workflow that runs on PR raised. As it's scanning the dependencies and ensuring only safe code is added.

So if you remove `dependency-review` you won't need this `if` conditional

AndyRae (changes requested):

AndyRae (approved):

Issue #1148, opened 2025-07-09
Add Airflow DAG for `search_recommendations` with Environment Toggle


brian-kim31:
### Is this the right issue type?

- [x] Yes, I'm planning work for this project team.

## Goal  
Implement a new Airflow DAG to generate mapping recommendations using basic string search against the OMOP concept table.

---

## 🔧 Configuration

- Add a new environment variable to control DAG execution:

  ```env
  SEARCH_ENABLED=false

## 🧩 DAG: `search_recommendations`

### Steps

1. Retrieve scan report value strings.
2. For each value string:
   - Search the OMOP concept table using:
     ```sql
     WHERE concept_name ILIKE '%{value_string}%'
     ```
   - Keep the search logic simple (no advanced fuzzy matching for now).
3. Select the first 3 results.
4. Insert them as new `MappingRecommendations` into the database.

---

## 🔁 Triggering the DAG

- Investigate and implement how to **trigger this DAG from another DAG**.
- This DAG should support being invoked as a **downstream dependency**.
- DAG execution should be **conditional based on the `SEARCH_ENABLED` environment variable**.

---

## ✅ Tasks

- [ ] Add support for `SEARCH_ENABLED` environment variable  
- [ ] Implement `search_recommendations` DAG  
- [ ] Use `ILIKE` query for concept matching  
- [ ] Select and store top 3 matches  
- [ ] Insert matches into `MappingRecommendations` table  
- [ ] Allow triggering from another DAG  
- [ ] Add documentation and examples  
- [ ] Tests

### Confirm creation

- [x] Ready to create this feature?

AndrewThien:
I wonder: 
- Since we can have the `vocab_id` from Data Dictionary, should we use this and other filters (`Standard` and `Valid` concepts) as well as `value_string` to improve the search
- Can `iLike` find an exact match as well? If so, can we have some conditions like:  
       - If we can use this search and find something 100% matching --> the search will stop --> the issue #983 solved
       - if nothing exactly matches, then we save these as recommendations  (I think it will then be something like `recommend by string matching`)
     
Is that right? @AndyRae

PR #1155, opened 2025-07-11
Mapping models

### LINKED ISSUE #1147: Add Database Models for MappingRecommendation and MappingTool

### Summary

We need to implement two new database models to support mapping recommendations generated by scan reports.

---

#### 1. `MappingRecommendation`

| Field               | Type                                                    | Description                                      | Nullable |
|---------------------|---------------------------------------------------------|--------------------------------------------------|----------|
| `id`                | `AutoField` (Primary Key)                               | Unique ID for the recommendation                 | False    |
| `created_at`        | `DateTimeField`                                         | Timestamp of when the recommendation was created | False    |
| `scan_report_value` | `GenericRelation(ScanReportValue, ScanReportField)`     | Link to scan report data                         | False    |
| `concept_id`        | `ForeignKey(Concept)`                                   | The concept this recommendation maps to          | False    |
| `score`             | `FloatField`                                            | Confidence or score of the recommendation        | True     |
| `tool`              | `ForeignKey(MappingTool)`                               | Tool used to generate the recommendation         | False    |

---

#### 2. `MappingTool`

| Field       | Type        | Description              | Nullable |
|-------------|-------------|--------------------------|----------|
| `id`        | `AutoField` | Unique ID                | False    |
| `name`      | `CharField` | Name of the tool         | False    |
| `version`   | `CharField` | Version of the tool      | False    |

---

### ✅ Tasks

- [x] Create the `MappingTool` model
- [x] Create the `MappingRecommendation` model
- [x] Add migrations
- [x] Register models in the admin panel 
- [ ] Tests pass.

### Confirm creation

- [x] Ready to create this feature?

brian-kim31:
✨ Feature

## PR Description
Added two new models to support mapping tool tracking and AI-powered mapping recommendations for carrot-mapper. The MappingRecommendation model uses Django's generic foreign key system (content_type and object_id) to enable flexible relationships with different content types, allowing recommendations to be attached to both ScanReportField and ScanReportValue objects without requiring separate foreign key fields.
The migration also seeds an initial "Search" mapping tool (version 1.0) to ensure there's a default tool available for existing functionality that may reference mapping tools, preventing potential foreign key constraint issues.

## Related Issues
Related #1147
Closes #1147


AndrewThien (commented):
I have one question:

Are entries in the MappingTool table required to be created manually by admins, before receiving future `MappingRecommendations`?

For example, by the seeding in the migration, we have one MappingTool with the id 1. In the future, if I want to get the mapping recommendation from Lettuce, I have to create the entries "Lettuce", with id 2 in the table MappingTool, before sending requests and receiving any responses from Lettuce, right? (otherwise, we can't fill the `tool_id` for the `MappingRecommendation` entries)

app/shared/shared/mapping/models.py
```diff
@@ -468,3 +470,38 @@ class Meta:
...
+
+
+class MappingRecommendation(BaseModel):
+    # These two fields are required for a generic relation
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    object_id = models.PositiveIntegerField()
+    scan_report_value = GenericForeignKey("content_type", "object_id")
+
+    concept = models.ForeignKey(Concept, on_delete=models.CASCADE)
+    score = models.FloatField()
```
AndyRae, line 503: In the first instance `score` should be `null=True, blank=True)`, as not everything will have a Score.

AndyRae (changes requested):

AndyRae:
> * Admins have to add the entries manually into the table across instances before Carrot can implement a new tool. This seems a bit unreasonable.
The datamodel here is to capture the origin of the search, it alone doesn't describe that it's the responsibility of an admin user to manage this?

> How can I manage the version of the tool? Shouldn't we get this directly from the response and add it directly to the MappingRecommendation record?
Yes that's the intention - this datamodel will capture that. But #1148 will populate the db with data from the tool and version used, if that version already exists - it will FK to it. If it doesn't it will create it.

AndrewThien:
> > * Admins have to add the entries manually into the table across instances before Carrot can implement a new tool. This seems a bit unreasonable.
> 
> The datamodel here is to capture the origin of the search, it alone doesn't describe that it's the responsibility of an admin user to manage this?
> 
> > How can I manage the version of the tool? Shouldn't we get this directly from the response and add it directly to the MappingRecommendation record?
> 
> Yes that's the intention - this datamodel will capture that. But #1148 will populate the db with data from the tool and version used, if that version already exists - it will FK to it. If it doesn't it will create it.

Okay, I see. But this means Airflow will need to check and compare the name and version between the current data and the incoming data every time, then if something new, it will create and use the new id? seems a bit not straightforward to me, compared to geting directly from the response, which will need to have the name and version anyway.

app/shared/shared/mapping/models.py
```diff
@@ -315,6 +315,7 @@ class ScanReportField(BaseModel):
     pass_from_source = models.BooleanField(default=True)
     field_description = models.CharField(max_length=256, blank=True, null=True)
     concepts = GenericRelation(ScanReportConcept)
+    mapping_recommendations = GenericRelation("MappingRecommendation")
```
AndrewThien, line 318: I think if you can move the MappingRecommendation class definition up before the class ScanReportField, then you don't need to have the quote here (and the consistency as on line 317 can be kept)

app/shared/shared/mapping/models.py
```diff
@@ -468,3 +470,38 @@ class Meta:
...
+
+    def __str__(self):
+        return f"{self.name} v{self.version}"
+
+
+class MappingRecommendation(BaseModel):
+    # These two fields are required for a generic relation
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    object_id = models.PositiveIntegerField()
+    scan_report_value = GenericForeignKey("content_type", "object_id")
```
AndrewThien, line 500: just wonder: why do you have here the field `scan_report_value`, but then have the `GenericRelation(MappingRecommendation)` in both `scan_report_value` and `scan_report_field` models

brian-kim31: I could name it better e.g content_object or related_object. I used GenericRelation to enable mapping recommendations at both field-level and value-level granularity, allowing AI tools to suggest mappings for entire fields or specific field values.

Also please see the whiteboard for details

https://whiteboard.cloud.microsoft/me/whiteboards/p/c3BvOmh0dHBzOi8vdW5pb2Zub3R0bS1teS5zaGFyZXBvaW50LmNvbS9wZXJzb25hbC9hbmR5X3JhZV9ub3R0aW5naGFtX2FjX3Vr/b!mAc72Dvk6UuVmxWuetQtl8zmGUHZCmRMuk2mkR3EwGnrX1IFgGvfTomSDAHImloJ/01BCY66MG3PCKAAD2CMNA3URUOY2TBD4WJ?fromShare=true

app/shared/shared/mapping/admin.py
```diff
@@ -203,6 +204,24 @@ class ProjectAdmin(admin.ModelAdmin):
...
+    list_display = (
+        "id",
+        "created_at",
+        "get_content_object",
+        "concept",
+        "score",
+        "tool_name",
+        "tool_version",
+    )
+    raw_id_fields = ("concept",)
```
brian-kim31, line 217: @AndrewThien 

I use raw_id_fields = ("concept",) because loading thousands of medical concepts into a dropdown would make the admin interface slow, so instead we show a simple ID input with a search popup for better performance.

PR #1161, opened 2025-07-24
Fix: Consistent File Size Limits for Scan Reports and Data Dictionaries

### LINKED ISSUE #1104: Make the body size limit for uploading consistent

### Summary

Currently, there are differences between the size limit when uploading SR and DD:

- No size limit for DD in backend
- Limit size for SR is 10mb, set by `DATA_UPLOAD_MAX_MEMORY_SIZE` in Backend
- Limit size for DD and SR is 20mb, set by `BODY_SIZE_LIMIT` in Frontend

These size limit should be consistent for both SR and DD in both backend and frontent to avoid slient upload failure/confusion.

### Acceptance Criteria

- [ ] size limit and validation consistent for both SR and DD in BE and FE 

### Tasks

- [ ] Task A.
- [ ] Task B.
- [ ] Unit tests.
- [ ] Documentation.

### Confirm creation

- [x] This issue is ready

brian-kim31:
⚡️ Optimization
## PR Description
This PR resolves inconsistencies in file size limits between Scan Reports (SR) and Data Dictionaries (DD) uploads, eliminating silent upload failures and improving user experience.

## Key Features Added
1. Consistent 20MB limit for both SR and DD files
2. Immediate user feedback when selecting oversized files
3. Clear error messages showing exact file sizes

Related #1104
Closes #1104

Screenshot 2025-07-24 at 16 49 15


AndrewThien:
I recommend you use `yup` package, which is already installed in Mapper, to validate the file size and show the errors on UI, instead of creating a custom validator.

app/next-client-app/components/scanreports/CreateScanReportForm.tsx
```diff
@@ -27,6 +28,50 @@ interface FormData {
   Data_dict: File | null;
 }
 
+// Yup validation schema
+const validationSchema = Yup.object({
```
AndrewThien, line 32: I don't think we need validations for things that already have their built-in validations. For example, `name` already has `required` by the Formik field or file types already have `accept=...`.

app/next-client-app/components/scanreports/CreateScanReportForm.tsx
```diff
@@ -245,37 +314,52 @@ export function CreateScanReportForm({
...
+                      <Input
+                        type="file"
+                        accept=".xlsx"
+                        required={true}
+                        onChange={(e) => {
+                          if (
+                            e.currentTarget.files &&
+                            e.currentTarget.files[0]
+                          ) {
+                            const file = e.currentTarget.files[0];
```
AndrewThien, line 339: I don't think we need one more validation for file size here. Similar for `Data_dict`.
So if users attempt to upload something too big, the <ErrorMessage> would be enough already

app/next-client-app/components/scanreports/CreateScanReportForm.tsx
```diff
@@ -245,37 +314,52 @@ export function CreateScanReportForm({
...
+                />
               </div>
 
               <div className="flex flex-col gap-2">
                 <h3 className="flex">
                   <div className="flex items-center gap-2">
                     Data Dictionary{" "}
                     <span className="text-muted-foreground text-sm">
-                      (.csv file, optional)
+                      (.csv file, optional, max 20MB)
```
AndrewThien, line 362: I wonder if you can make the number here configurable as well? So I think you can try putting the `bodySizeLimit` into the constants/index.ts, then import it here and `next.config.ts`.
Because I don't want this number fixed here, in case we want to modify the size limit flexibly

app/next-client-app/components/scanreports/CreateScanReportForm.tsx
```diff
@@ -27,6 +28,50 @@ interface FormData {
...
+const validationSchema = Yup.object({
+  name: Yup.string().required("Scan Report name is required"),
+  dataPartner: Yup.number().min(1, "Please select a Data Partner"),
+  dataset: Yup.number().min(1, "Please select a Dataset"),
+  scan_report_file: Yup.mixed()
+    .required("Scan Report file is required")
+    .test("fileSize", "File size must be less than 20MB", function (value) {
+      if (!value) return true;
+      const file = value as File;
+      return file.size <= 20 * 1024 * 1024;
```
AndrewThien, line 41: this number should be configurable. Again, try putting the size limit in one place then use it here

AndrewThien (changes requested):
can you please set the default size limits in both BE and FE to 30MB? we had one request from a user to increase this limit. Thanks!

app/next-client-app/components/scanreports/CreateScanReportForm.tsx
```diff
@@ -99,12 +145,20 @@ export function CreateScanReportForm({
...
           toast.info("Validating ...");
           handleSubmit(data);
         }}
       >
-        {({ values, handleChange, handleSubmit, setFieldValue }) => (
+        {({
+          values,
+          handleChange,
+          handleSubmit,
+          setFieldValue,
```
AndrewThien, line 158: do we need or use `setFieldValue` and `touched` for something here?

AndrewThien (changes requested):
I think you tested the PR before marking it ready for review, but you only tested uploading both SR and DD. In production, users can upload only SR, so please take this into account.

AndrewThien (dismissed):

AndrewThien:
I tried to import stuff to nextjs.config. It is impossible but we would need to use `require` pattern, not `import`. Also tried other ways. And I think the best way for now is to make the variable become `NEXT_PUBLIC_BODY_SIZE_LIMIT` then use `process.env.NEXT_PUBLIC_BODY_SIZE_LIMIT` to access it from nextjs.config and client component. This would prevent the overhead of having two env variables.

However, if we choose this approach, we need to bake the NEXT_PUBLIC_BODY_SIZE_LIMIT into the GH workflow when deploying NextJS app

AndrewThien:
on the second thought, using `require` would be an okay option as well. So we can put the env var into the constants file, then `import` to client component and `require` to nextjs.config

app/next-client-app/constants/index.ts
```diff
@@ -19,3 +19,8 @@ export const enableAIRecommendation = env(
 export const recommendationServiceName = env(
   "NEXT_PUBLIC_RECOMMENDATION_SERVICE_NAME"
 );
+
+// File size limit in bytes (configurable via env, default 30MB)
+export const MAX_FILE_SIZE_BYTES = process.env.NEXT_PUBLIC_BODY_SIZE_LIMIT
```
AndyRae, line 24: This is also defined and exported in `config.js` as well, so this definition is likely not needed?

AndyRae: Yes - it's the only way we can import into `next.config.js` using `require` yes?

brian-kim31: Yes its the only way

AndyRae (approved):

PR #1165, opened 2025-07-30
Added Show users on datasets/scan reports

### LINKED ISSUE #1073: Show users on datasets/scan reports

### Summary

Projects have the ability to see who has access:
Image

This would be good to add to datasets, and scan reports also.

### Acceptance Criteria

- [ ] Users can be seen on datasets and scan reports
- [ ] Tests pass.

### Tasks

- [ ] Add to datasets
- [ ] Add to scan reports

### Confirm creation

- [x] This issue is ready

brian-kim31:
✨ Feature

## PR Description
This PR implements member display functionality across datasets and scan reports by creating dedicated API functions (getDatasetMembers() and getScanReportMembers()) that aggregate and deduplicate users from multiple permission levels (viewers, editors, and admins for datasets; viewers and editors for scan reports). The implementation integrates with existing layout components to show all users with access to each resource using the AvatarList component, which displays member avatars with tooltips and a count indicator. This provides users with immediate visibility into who has access to datasets and scan reports, improving collaboration and access management while maintaining the existing permission structure and ensuring backward compatibility.

Related #1073
Closes #1073

Screenshot 2025-08-01 at 12 35 45
Screenshot 2025-08-01 at 12 35 45

### DIFF
--- a/app/airflow/dags/libs/rules_export/core.py
+++ b/app/airflow/dags/libs/rules_export/core.py
@@ -2,14 +2,18 @@ import logging
 from libs.utils import pull_validated_params
 from airflow.providers.postgres.hooks.postgres import PostgresHook
 from libs.types import FileHandlerConfig
-from libs.rules_export.file_services import build_rules_json, build_rules_csv
+from libs.rules_export.file_services import (
+    build_rules_json,
+    build_rules_csv,
+    build_rules_json_v2,
+)
 from typing import Dict
 from datetime import datetime
 from libs.queries import create_update_temp_rules_table_query, create_file_entry_query
 from libs.storage_services import upload_blob_to_storage
 from libs.enums import JobStageType, StageStatusType
 from libs.utils import update_job_status
-from libs.settings import AIRFLOW_DEBUG_MODE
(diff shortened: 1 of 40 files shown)


AndyRae:
I can see why this looks like a small issue to tackle, and why you'd take this approach given the existing codebase and how it fetches / resolves this pattern generally. But this is really an iceberg issue - looks small, but there's a bunch of debt it's building on.

The ideal pattern for this is to mirror the existing `Projects` type / API, when fetching a Project the API contains the list of members (as user objects), this makes it easy for the frontend to render it.

Unfortunately - the pattern is not the case for `Datasets` and `Scan Reports, where as you've seen only serialize the user ids on their endpoints, so therefore you need to go and fetch the users again, and we're back to adding more code on top of a bad pattern.

So what do to?
1. Merge this (layer on top of the bad pattern and make it harder to fix in future)
2. Fix the pattern and rework this slightly
3. Leave it for now - this is a P3 issue and we have bigger fish to fry! This is my current thinking but happy for thoughts @AndrewThien @brian-kim31

AndyRae (approved):

Issue #1075, opened 2025-05-23
Forms improvements


AndyRae:
### Is this the right issue type?

- [x] Yes, I'm planning work for this project team.

### Summary

The forms on Carrot are generally annoying.

For example:

<img width="1440" alt="Image" src="https://github.com/user-attachments/assets/cc56044f-5734-4595-a724-a0cd7785acd3" />

- Too wide width
- Name field doesn't accept edits, bug with state here I think
- Dropdowns should be ordered alphabetically (not ID?)
- #1070 
- Icons should be captions
- Spacing between rows should be higher
- Save button is really ugly? Doesn't match other buttons in the app

Good design: https://ui.shadcn.com/examples/forms

### Acceptance Criteria

- [x] Forms have been improved
- [x] Tests pass.

### Tasks

- [x] Reduce width
- [x] Name field fix state
- [x] Dropdowns order fix
- [x] Move icons to captions
- [x] Increase spacing between rows
- [x] Fix buttons + icons
- [x] Forms when submitted should redirect to logical places

Forms
- [x] Upload Scan Report
- [x] "Edit Table"
- [x] Edit Scan Report Details
- [x] Edit Dataset Details
- [x] Update field

### Confirm creation

- [x] This issue is ready

PR #1167, opened 2025-07-31
Update forms for consistency and improved experience

AndyRae:
| |
|-|
♻️ Refactor
✨ Feature

## PR Description
Updates the apps forms for a consistent UI and improved user experience.

The forms have been rebuilt using the shadcn components that have been reduced back to their original styles, with a focus on providing consistent forms experience across the entire application.

This is a continuation made possible by #1097.

I tried to keep this limited to just the forms, but it does creep into other components slightly.. so this work continued in #1166 

Related #1075 

Screenshot 2025-07-31 at 20 36 33
Screenshot 2025-07-31 at 20 28 43
Screenshot 2025-07-31 at 20 28 50
Screenshot 2025-07-31 at 20 29 04
Screenshot 2025-07-31 at 20 29 30


AndyRae:
> Should have given comments on this earlier. But I just wondered if it would be nicer when the upload SR form is in the middle of the page? Cause this page only has this form and the menu bar.

Really happy for more changes - on this one specifically, interestingly I had the same thought. The reason I didn't is the Scan Reports breadcrumb at the top, which felt suddenly in the middle of nowhere. I then figured, kind of odd to have one piece of centre aligned when nothing else is.

But yeah - I think it could work with a bit of thought

PR #1227, opened 2025-11-26
Enhanced scan report upload speed

### LINKED ISSUE #1226: Optimize bulk inserts in scan report processing to prevent timeouts

### Current Behavior

The update_temp_data_dictionary_table() and create_temp_field_values_table() functions use pg_hook.run() in loops to insert records one at a time. For large scan reports with thousands of records, this causes:

1. Slow performance: Each record requires a separate database round trip, resulting in execution times of several minutes for large datasets
2. DAG timeouts: The scan report processing DAG has a 60-minute timeout, and bulk inserts can exceed this limit, causing DAG runs to fail
3. Inefficient database usage: Multiple individual INSERT statements create unnecessary overhead (parsing, planning, transaction management for each row)

Affected Functions

- update_temp_data_dictionary_table() in app/airflow/dags/libs/SR_processing/db_services.py
- create_temp_field_values_table() in app/airflow/dags/libs/SR_processing/db_services.py

### Expected Behavior

Bulk inserts should use psycopg2.extras.execute_values() to batch all records into a single INSERT statement with multiple VALUES. This should:

1. Dramatically improve performance: Insert thousands of records in seconds instead of minutes
2. Prevent timeouts: Complete bulk insert operations well within the DAG timeout window
3. Reduce database overhead: Single round trip and transaction for all records

The fix should maintain the same functionality and error handling, but with significantly better performance for large datasets.

### Steps To Reproduce

Uploading a large scan report and data dictionary. Below are the exact files that exposed the issue:

[DataDictionary_Nov_25.csv](https://github.com/user-attachments/files/23776096/DataDictionary_Nov_25.csv)
[ScanReport_Nov_25.xlsx](https://github.com/user-attachments/files/23776097/ScanReport_Nov_25.xlsx)

_No response_

### Anything else?

_No response_

brian-kim31:
⚡️ Optimization

## PR Description

The bulk inserts in update_temp_data_dictionary_table() and create_temp_field_values_table() were using pg_hook.run() in loops, which meant one INSERT per row. For large scan reports with thousands of records, this was slow and could cause DAG timeouts.
This PR switched to psycopg2.extras.execute_values() to batch all rows into a single INSERT, which is much faster and avoids timeout issues.

Related #1226
Closes #1226

Below is a screenshot of the new record time it takes to upload the scan report and data dictionary

Screenshot 2025-11-26 at 17 54 53

### DIFF
--- a/app/airflow/dags/libs/SR_processing/db_services.py
+++ b/app/airflow/dags/libs/SR_processing/db_services.py
@@ -125,7 +125,7 @@ def update_temp_data_dictionary_table(
                     "value_description",
                 ],
                 fast_executemany=True,
-                commit_every=3000,
+                commit_every=10000,
             )
 
         logging.info(
@@ -214,7 +214,7 @@ def create_temp_field_values_table(
                     "frequency",
                 ],
                 fast_executemany=True,
-                commit_every=3000,
+                commit_every=10000,
             )
 
     except Exception as e:
--- a/app/airflow/dags/libs/SR_processing/db_services.py
+++ b/app/airflow/dags/libs/SR_processing/db_services.py
@@ -5,9 +5,14 @@ from typing import Any, Dict, List, Tuple
 from airflow.providers.postgres.hooks.postgres import PostgresHook
 from libs.enums import JobStageType, StageStatusType
 from libs.queries import create_fields_query
-from libs.settings import AIRFLOW_DAGRUN_TIMEOUT, AIRFLOW_DEBUG_MODE
+from libs.settings import (
+    AIRFLOW_DAGRUN_TIMEOUT,
+    AIRFLOW_DEBUG_MODE,
+    EXECUTE_VALUES_PAGE_SIZE,
+)
 from libs.utils import update_job_status
 from openpyxl.worksheet.worksheet import Worksheet
+from psycopg2.extras import execute_values
 
 # PostgreSQL connection hook
 pg_hook = PostgresHook(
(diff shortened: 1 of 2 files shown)


app/airflow/dags/libs/SR_processing/db_services.py
```diff
@@ -105,26 +106,35 @@ def update_temp_data_dictionary_table(
...
-                    "table_name",
-                    "field_name",
-                    "value",
-                    "value_description",
-                ],
-            )
+            conn = pg_hook.get_conn()
+            cursor = conn.cursor()
+            try:
+                execute_values(
```
prquinlan: It will be good to have some sanity check and batching on this input, at the moment I think this could be any size and we probably want a batch of a certain size. 

Previously pg_hook.insert_rows did this automatically so we would want something similar to ensure it is robust.

prquinlan (changes requested):
Comments added to ensure we have a mechanism to manage large inputs and to have a batching strategy.

app/airflow/dags/libs/SR_processing/db_services.py
```diff
@@ -195,23 +205,34 @@ def create_temp_field_values_table(
...
-                target_fields=[
-                    "field_name",
-                    "value",
-                    "frequency",
-                ],
-            )
+            conn = pg_hook.get_conn()
+            cursor = conn.cursor()
+            try:
+                execute_values(
```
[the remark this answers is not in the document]
AndyRae: Ah that's interesting, what do you mean by it's not working? 

I'm surprised as this is from the Postgres provider documentation: https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/_api/airflow/providers/postgres/hooks/postgres/index.html

And is very much in the Postgres code: https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/_modules/airflow/providers/postgres/hooks/postgres.html#PostgresHook.insert_rows

And in psycopg2: https://www.psycopg.org/docs/extras.html#fast-execution-helpers

Do you have an error message? 

There is also more than `fast_executemany` to tune - so understanding what is not working before we drop down would be good.

AndyRae:
> On trying with a commit of 3000, it took 59 minutes. No dramatic change. I'm gonna try with 6000 and see the result.
> 

That is interesting - what do the Airflow logs say they are running on the database?

app/airflow/dags/libs/SR_processing/db_services.py
```diff
@@ -124,6 +124,8 @@ def update_temp_data_dictionary_table(
                     "value",
                     "value_description",
                 ],
+                fast_executemany=True,
+                commit_every=0,  # commit every row to avoid transaction overhead
```
AndyRae: It'd be helpful to have this as an environment variable in config, so we tune through configuration rather than deployment.

AndyRae (changes requested):

